PR #8, opened 2023-07-24
Parse function contracts

zzqatuber:
This PR supports parsing line comments before function declarations as contracts.

### Design

The line comment must contain one or more contracts but nothing else. The following two comments are not parsed as function contracts.
```go
// if we have `// contract(nonnil->nonnil)`
foo(...)...
```
```go
// ...when we have `//
// contract(nonnil->nonnil)`.
foo(...)...
```
A single contract must follow synatx as follows:
```
contract(VALUE[,VALUE]+ -> VALUE[,VALUE]+)
```
`VALUE` must be one of `nonnil`, `true`, `false`, `_`, and whitespaces are allowed between any two tokens.

### Implementation

Create a new analyzer that is a prerequisite of the existing function analyzer.

### Test Plan
Functional:
- A new analyzer_test is added.
- Sanity checking CI job should pass.

Performance:
- Tested in the last pull request of the stack, i.e., nonnil->nonnil no infer mode.

### DIFF
--- a/assertion/function/analyzer.go
+++ b/assertion/function/analyzer.go
@@ -28,6 +28,7 @@ import (
 	"go.uber.org/nilaway/annotation"
 	"go.uber.org/nilaway/assertion/anonymousfunc"
 	"go.uber.org/nilaway/assertion/function/assertiontree"
+	"go.uber.org/nilaway/assertion/function/functioncontracts"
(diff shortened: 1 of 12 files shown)


yuxincs (approved):
This PR has been reviewed internally and all comments have been addressed. Let's wait until CI is working before we merge 😃

PR #10, opened 2023-07-26
Replace config.IsTestEnvironment with a configuration

yuxincs:
This PR replaces the global variable `config.IsTestEnvironment` with a user-configuration option `pretty-print` (default `true`). Then in our `TestMain` setup function we set this flag to be false for testing purposes.

Depends on #9


lazaroclapp (commented):
Overall looks good, but do we want a test specifically to test that setting this flag has some effect, even if we disable it for the majority of the test suite?

yuxincs:
> Overall looks good, but do we want a test specifically to test that setting this flag has some effect, even if we disable it for the majority of the test suite?

Right, it's probably better to test it. I have added a non-parallel test to the test suite to test this functionality (since it needs to change the global `config.Analyzer`, and then revert it for the other tests).

lazaroclapp (approved):
LGTM!

nilaway_test.go
```diff
@@ -216,9 +217,25 @@ func TestGenerics(t *testing.T) {
 	analysistest.Run(t, testdata, Analyzer, "go.uber.org/generics")
 }
 
+func TestPrettyPrint(t *testing.T) { //nolint:paralleltest
+	// We specifically do not set this test to be parallel such that this test is run separately
+	// from the parallel tests. This makes it possible to set the pretty-print flag to true for
+	// testing and false for the other tests.
```
yuxincs, line 223: This is also documented here: https://pkg.go.dev/testing#T.Parallel

PR #11, opened 2023-07-26
Remove hardcoded file ignore string check

yuxincs:
This PR removes the hardcoded `nolint:nilaway` docstring check for both files and function declarations since it is no longer necessary:

* files: we have implemented configuration analyzer in #9 which is able to take a list of strings to ignore the analysis of a file.

* function declaration: it is mainly designed for two reasons:
  * performance: we were able to use it to manually skip the analysis of a particular function for performance reasons. However, we now have a timeout for analysis of any particular function, hence this is no longer required.
  * error suppression: most linter drivers, especially nogo, are now respecting `nolint`, so there is really no need to support it within the linter itself.

Depends on #10


lazaroclapp (approved):
LGTM!

PR #18, opened 2023-08-03
Add logic for error suppression and passing configs in the standalone driver

### LINKED ISSUE #16: Nilaway execute detection recursively in all of the dependencies
I want an option that Nilaway only detects in a certain repo.

yuxincs:
This PR adds error suppression logic (since `singlechecker` does not support it) and passing configs (by lifting flags from `config.Analyzer` for better UX) to the standalone checker.

Now it is possible to specify flags for NilAway via

```
nilaway -include-pkgs=go.uber.org -include-errors-in-files=./ ./...
```
such that only packages starting with `go.uber.org` is _analyzed_, and only errors reported in the files within the current directory is _reported_.

Please note the differences between _analyzed_ and _reported_: even if a package is excluded from analysis, errors might still be reported on them if the nilness flows happens within the analyzed package, but it includes using a struct from the excluded package. For example, if the analyzed package constructs a struct from the excluded package, assigns a nil to a field, and then immediately dereferences the field. In this case, NilAway will report an error for this flow, but the location will be on the field (that resides in the excluded package).

By default, the standalone checker sets `-include-errors-in-files` to the current working directory, so NilAway won't report any errors on other packages. For other linter drivers, uses will have to set this through the drivers instead.

The [wiki/Configuration](https://github.com/uber-go/nilaway/wiki/Configuration) is also updated to reflect this change, and a link to the configuration page is added to the top-level README.

This PR also adds another linting task to run NilAway on itself.

Fixes #16.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ cover:
 	go tool cover -html=cover.out -o cover.html
 
 .PHONY: lint
-lint: golangci-lint tidy-lint
+lint: golangci-lint nilaway-lint tidy-lint
 
 .PHONY: golangci-lint
 golangci-lint:
@@ -39,3 +39,8 @@ tidy-lint:
 	@go mod tidy && \
 		git diff --exit-code -- go.mod go.sum || \
 		(echo "'go mod tidy' changed files" && false)
+
+.PHONY: nilaway-lint
+nilaway-lint: build
+	@echo "[lint] nilaway linting itself"
+	@$(GOBIN)/nilaway -include-pkgs="go.uber.org/nilaway" ./...
--- a/assertion/function/analyzer.go
+++ b/assertion/function/analyzer.go
@@ -130,7 +130,7 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
 	var funcIndex int
 	for _, file := range pass.Files {
 		// Skip if a file is marked to be ignored, or it is not in scope of our analysis.
-		if util.DocContainsIgnore(file.Doc) || !conf.IsFileInScope(file) {
+		if !conf.IsFileInScope(file) {
 			continue
 		}
 
@@ -189,8 +189,8 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
 				panic(fmt.Sprintf("unrecognized function type %T", f))
 			}
 
-			// Skip if function declaration has an empty body, or it is marked to be ignored.
-			if funcDecl.Body == nil || util.DocContainsIgnore(funcDecl.Doc) {
+			// Skip if function declaration has an empty body.
+			if funcDecl.Body == nil {
 				continue
 			}
 			// Skip if the function is too large.
--- a/assertion/function/analyzer.go
+++ b/assertion/function/analyzer.go
@@ -138,7 +138,7 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
 		// on NilAway itself.
 		functionConfig := assertiontree.FunctionConfig{}
 		if strings.HasPrefix(pass.Pkg.Path(), config.NilAwayPkgPathPrefix) { //nolint:revive
-			// TODO: enable struct initialization flag.
+			// TODO: enable struct initialization flag (tracked in Issue #23).
 			// TODO: enable anonymous function flag.
 		} else {
 			functionConfig.StructInitCheckType = util.DocContainsStructInitCheck(file.Doc)
--- a/assertion/function/analyzer.go
+++ b/assertion/function/analyzer.go
@@ -137,9 +137,9 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
 		// Construct config for analyzing the functions in this file. By default, enable all checks
 		// on NilAway itself.
 		functionConfig := assertiontree.FunctionConfig{}
-		if strings.HasPrefix(pass.Pkg.Path(), config.NilAwayPkgPathPrefix) {
-			functionConfig.StructInitCheckType = config.DepthOneFieldCheck
-			// TODO: enable anonymous function flag here when  is done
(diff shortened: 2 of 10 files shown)


Makefile
```diff
@@ -39,3 +39,8 @@ tidy-lint:
 	@go mod tidy && \
 		git diff --exit-code -- go.mod go.sum || \
 		(echo "'go mod tidy' changed files" && false)
+
+.PHONY: nilaway-lint
+nilaway-lint: build
+	@echo "[lint] nilaway linting itself"
```
lazaroclapp, line 45: 🎉

assertion/function/analyzer.go
```diff
@@ -137,9 +137,9 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
 		// Construct config for analyzing the functions in this file. By default, enable all checks
 		// on NilAway itself.
 		functionConfig := assertiontree.FunctionConfig{}
-		if strings.HasPrefix(pass.Pkg.Path(), config.NilAwayPkgPathPrefix) {
-			functionConfig.StructInitCheckType = config.DepthOneFieldCheck
-			// TODO: enable anonymous function flag here when  is done
+		if strings.HasPrefix(pass.Pkg.Path(), config.NilAwayPkgPathPrefix) { //nolint:revive
+			// TODO: enable struct initialization flag.
```
lazaroclapp, line 141: Why are we removing this now? Is it that we need to add it as another CLI flag for the stand alone mode? Is that hard to do?

cmd/nilaway/main.go
```diff
@@ -18,10 +18,105 @@
...
+			files[i] = nil
+			continue
+		}
+
+		// Convert the file paths to absolute paths.
+		list := strings.Split(flagVal, ",")
+		for i := range list {
+			p, err := filepath.Abs(list[i])
+			if err != nil {
+				return nil, fmt.Errorf("incorrect file path: %s", list[i])
```
lazaroclapp, line 73: Do we not want to include the underlying error `err` here somehow? Just to get a more specific error message.

cmd/nilaway/main.go
```diff
@@ -18,10 +18,105 @@
...
+	// NilAway by default analyzes all packages, including dependencies. Even if specified to
+	// exclude packages from analysis via configurations, NilAway can still report errors on
+	// packages that are not analyzed if the nilness flow happens within the analyzed package, but
+	// the flow concerns a struct that is in an excluded package. The usual way to handle them is
+	// to suppress them at the driver level, but singlechecker does not support that yet. Therefore,
+	// here we add extra logic to filter the errors.
+
+	// Properly parse the error suppression flags.
+	var files [2][]string
+	for i, flagVal := range [...]string{_includeErrorsInFiles, _excludeErrorsInFiles} {
```
lazaroclapp, line 62: Is this loop cleaner than extracting this logic into a `readFileList(...)` function or similar?

Then do:

```
includes, err := readFileList(_includeErrorsInFiles)
if err != nil {
  return nil, err
}
excludes, err2 := readFileList(_excludeErrorsInFiles)
if err != nil {
  return nil, err
}
```

Or something like that?

cmd/nilaway/main.go
```diff
@@ -18,10 +18,105 @@
...
+		if flagVal == "" {
+			files[i] = nil
+			continue
+		}
+
+		// Convert the file paths to absolute paths.
+		list := strings.Split(flagVal, ",")
+		for i := range list {
```
lazaroclapp, line 70: To avoid confusion, wouldn't it be better to use `j` than to shadow `i`? Or is there a Go style guideline rule against it?

build/tools
lazaroclapp, line 1: What is this and why is it a submodule? What does it point to? (Can't figure this out from the GH review interface)

cmd/nilaway/main.go
```diff
@@ -18,10 +18,105 @@
...
+	}
+
+	// Delegate the real analysis run to the original nilaway analyzer.
+	return nilaway.Analyzer.Run(pass)
+}
+
 func main() {
-	singlechecker.Main(nilaway.Analyzer)
+	// For better UX, we lift the flags from config.Analyzer to the top level so that users can
+	// specify them without having to specify the analyzer name ("nilaway_config").
```
lazaroclapp, line 105: Sorry, not sure I understand what this is doing. Does this mean we can pass `nilaway:flagName` in the command line and it will be passed to `nilaway_config`, or just `--flagName`? Can you give an example here on how this looks from the user point of view?

assertion/function/analyzer.go
yuxincs, line 141: We had to disable struct init for NilAway itself due to too many false postives (especially related to escaped fields). We should investigate this in the future and enable it.

lazaroclapp: Do we have an issue tracking this in GH? If so, let's add the issue number on the TODO comment.

cmd/nilaway/main.go
yuxincs: Without this, users will have to understand our internal architecture and pass the flags specifically to the "nilaway_config" analyzer:

```
nilaway -nilaway_config.flag1 <VALUE1> -nilaway_config.flag2 <VALUE2>
```

With the lifting logic here, this will become transparent to the users:

```
nilaway -flag1 <VALUE1> -flag2 <VALUE2>
```

and internall we pass those flags to the "nilaway_config" analyzer. 

I have updated the comments here to include this example (hopefully it makes it clearer)

build/tools
yuxincs: Ah....sorry, this is where I was storing some build artifacts and got in this PR (it is a submodule since `tools` is a git repo, `go/tools` for testing NilAway). 

I have added `build/` to gitignore so that it won't get included in the future :)

lazaroclapp (approved):
LGTM! Thanks for all the clarifications!

PR #20, opened 2023-08-04
Add codecov step in CI and add badges to README

yuxincs:
This PR adds a codecov step in CI configs and add the corresponding badges to README.

We have also added a config file for codecov to fail PRs if the coverage decreases (experimental, we can revert / disable it if it gets too annoying in the future)

### DIFF
--- a/.codecov.yaml
+++ b/.codecov.yaml
@@ -1,0 +1,14 @@
+coverage:
+  range: 80..100
+  round: down
+  precision: 2
+
+  status:
+    project:                   # measuring the overall project coverage
+      default:                 # context, you can create multiple ones with custom titles
+        enabled: yes           # must be yes|true to enable this status
+        target: auto           # specify the target coverage for each commit status
+                               #   option: "auto" (must increase from parent commit or pull request base)
+                               #   option: "X%" a static target percentage to hit
+        if_not_found: success  # if parent is not found report status as success, error, or failure
+        if_ci_failed: error    # if ci fails report status as success, error, or failure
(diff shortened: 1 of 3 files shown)


.codecov.yaml
```diff
@@ -0,0 +1,14 @@
...
+  range: 80..100
+  round: down
+  precision: 2
+
+  status:
+    project:                   # measuring the overall project coverage
+      default:                 # context, you can create multiple ones with custom titles
+        enabled: yes           # must be yes|true to enable this status
+        target: auto           # specify the target coverage for each commit status
+                               #   option: "auto" (must increase from parent commit or pull request base)
```
lazaroclapp, line 11:
```suggestion
                               #   option: "auto" (must not decrease from parent commit or pull request base)
```

Right?

lazaroclapp (approved):
🚀

.codecov.yaml
yuxincs: Right, actually I found out that we can do `auto` with a `threshold` https://docs.codecov.com/docs/commit-status and I have set this to be 5%, meaning that the coverage cannot drop more than 5% (to also allow some flakiness). 

Let me know if you think we should be stricter about this :)

lazaroclapp: SGTM!

lazaroclapp (approved):
Actually, can we make the threshold drop 1-2%. Given the size of the codebase, coverage going down by 5% can easily be a pretty huge PR with no tests. No need for me to re-stamp, though.

PR #21, opened 2023-08-04
Run CI builds on all PRs

yuxincs:
This PR changes the CI build configs such that CI is run on all PRs, instead of only on PRs that are against the `main` branch. This will allow us to have early CI feedback for stacked PRs.


lazaroclapp (approved):
SGTM! (As we discussed, this is all subject to external contributors needing us to kickstart their builds, but that's true regardless of which branch the PR is against :) )

PR #22, opened 2023-08-04
Fix producer expr nil

sonalmahajan15:
Producer expressions that are nil or do not point to the correct expression in the AST cause their position information to be not reported in the error message. This diff applies fixes to those cases.

Below is an example of a before and after error message:

```
7.   func testProducerPosPrint() {
8.	    var i *int
9. 	    print(*i)
10.  }
```

**Before:**
Value read from a variable that was never assigned to (definitely nilable) and is dereferenced at "errormessages/errormessages.go:9:9" (must be nonnil)

**After:**
Value read from a variable that was never assigned to at "errormessages/errormessages.go:8:6" (definitely nilable) and is dereferenced at "errormessages/errormessages.go:9:9" (must be nonnil)


.gitignore
```diff
@@ -26,3 +26,6 @@ cover.html
 
 # Local binary folder
 bin/
+
+# Mac OS files
+**/.DS_Store
```
lazaroclapp, line 31: I think this file doesn't end on a newline character. Can we make sure it does?

util/util.go
```diff
@@ -213,8 +213,8 @@ func PortionAfterSep(input, sep string, occ int) string {
 // Right now, this is used only to decide whether to print the location of the producer expression
 // in a full trigger.
 func ExprIsAuthentic(pass *analysis.Pass, expr ast.Expr) bool {
-	_, ok := pass.TypesInfo.Types[expr]
-	return ok
+	t := pass.TypesInfo.TypeOf(expr)
+	return t != nil
```
lazaroclapp, line 217: This one is also an unrelated fix, right? Looks good to me, just confirming it hasn't much to do with the PR goal.

lazaroclapp (approved):
A quick question and a `\n` requested below, but otherwise LGTM 🙂

.gitignore
yuxincs: We could probably get some formatters in the linting task to automatically do that. Let me try to find a good one (or do you have anything in mind?)

yuxincs (approved):
LGTM apart from Lazaro's question on whether a fix is related or not.

util/util.go
sonalmahajan15: No, this is actually related a fix. The original [TypesInfo.Types](https://pkg.go.dev/go/types#Info.Types) does not capture all identifiers, such as `var i *int`. While, [TypesInfo.TypeOf](https://pkg.go.dev/go/types#Info.TypeOf) comprehensively captures all expressions.

So, in the simple code example below, position for uninitialized variable `i` was not being reported earlier, which works fine now.

```
var i *int
print(*i)
```

lazaroclapp (approved):
LGTM!

PR #24, opened 2023-08-04
[Follow-up of PR #8 ] Support function contract nonnil->nonnil for full infer mode

zzqatuber:
Follow up of PR #8 , support function contract(nonnil->nonnil) in full infer mode.

This revision supports contract(nonnil -> nonnil). There are two major steps:

1. Duplicate param and return sites at every call of the function with a contract so a param or return site will have its unique nilability. We duplicate full triggers from the contracted functions as well in order those duplicated param and return sites are involved in those duplicated full triggers.

2. Introduce controlled triggers to allow NilAway to ignore some over-constraints when the contract should work.

### DIFF
--- a/annotation/consume_trigger.go
+++ b/annotation/consume_trigger.go
@@ -500,8 +500,8 @@ func (m MethodParamFromInterfacePrestring) String() string {
 		m.ParamName, m.ImplName, m.IntName)
 }
 
-// DuplicateReturnConsumer duplicates a given consumer trigger, assuming the given consumer trigger
-// is of UseAsReturn.
+// DuplicateReturnConsumer duplicates a given consume trigger, assuming the given consumer trigger
+// is for a UseAsReturn annotation.
 func DuplicateReturnConsumer(t *ConsumeTrigger, location token.Position) *ConsumeTrigger {
 	ann := t.Annotation.(UseAsReturn)
 	key := ann.TriggerIfNonNil.Ann.(RetAnnotationKey)
--- a/annotation/consume_trigger.go
+++ b/annotation/consume_trigger.go
@@ -19,6 +19,7 @@ import (
 	"go/ast"
 	"go/token"
 	"go/types"
+	"strings"
 
 	"go.uber.org/nilaway/util"
 )
@@ -355,7 +356,11 @@ func (g GlobalVarAssignPrestring) String() string {
 	return fmt.Sprintf("assigned into the global variable `%s`", g.VarName)
 }
 
-// ArgPass is when a value flows to a point where it is passed as an argument to a function
+// ArgPass is when a value flows to a point where it is passed as an argument to a function.
+// This consumer trigger can be used on top of two different sites: ParamAnnotationKey &
+// ArgAnnotationKey. ParamAnnotationKey is the parameter site in the function declaration;
+// ArgAnnotationKey is the argument site in the call expression. ArgAnnotationKey is specifically
+// used for functions with contracts since we need to duplicate the sites for context sensitivity.
 type ArgPass struct {
 	TriggerIfNonNil
 }
@@ -366,10 +371,21 @@ func (a ArgPass) String() string {
 
 // Prestring returns this ArgPass as a Prestring
 func (a ArgPass) Prestring() Prestring {
-	paramAnn := a.Ann.(ParamAnnotationKey)
-	return ArgPassPrestring{
-		ParamName: paramAnn.MinimalString(),
-		FuncName:  paramAnn.FuncDecl.Name(),
+	switch key := a.Ann.(type) {
+	case ParamAnnotationKey:
+		return ArgPassPrestring{
+			ParamName: key.MinimalString(),
+			FuncName:  key.FuncDecl.Name(),
+			Location:  "",
+		}
+	case ArgAnnotationKey:
+		return ArgPassPrestring{
+			ParamName: key.MinimalString(),
+			FuncName:  key.FuncDecl.Name(),
+			Location:  key.Location.String(),
+		}
+	default:
+		panic(fmt.Sprintf("Expected ParamAnnotationKey or ArgAnnotationKey but see: %T", key))
 	}
 }
 
@@ -377,10 +393,18 @@ func (a ArgPass) Prestring() Prestring {
 type ArgPassPrestring struct {
 	ParamName string
 	FuncName  string
+	// Location points to the code location of the argument pass at the call site for a ArgPass
+	// enclosing ArgAnnotationKey; Location is empty for a ArgPass enclosing ParamAnnotationKey.
+	Location string
 }
 
 func (a ArgPassPrestring) String() string {
-	return fmt.Sprintf("passed as %s to func `%s`", a.ParamName, a.FuncName)
+	var sb strings.Builder
+	sb.WriteString(fmt.Sprintf("passed as %s to func `%s`", a.ParamName, a.FuncName))
+	if a.Location != "" {
+		sb.WriteString(fmt.Sprintf(" at %s", a.Location))
+	}
+	return sb.String()
 }
 
 // RecvPass is when a receiver value flows to a point where it is used to invoke a method.
@@ -474,7 +498,29 @@ func (m MethodParamFromInterfacePrestring) String() string {
 		m.ParamName, m.ImplName, m.IntName)
 }
 
-// UseAsReturn is when a value flows to a point where it is returned from a function
+// DuplicateReturnConsumer duplicates a given consumer trigger, assuming the given consumer trigger
+// is of UseAsReturn.
+func DuplicateReturnConsumer(t *ConsumeTrigger, location token.Position) *ConsumeTrigger {
+	ann := t.Annotation.(UseAsReturn)
+	key := ann.TriggerIfNonNil.Ann.(RetAnnotationKey)
+	return &ConsumeTrigger{
+		Annotation: UseAsReturn{
+			TriggerIfNonNil: TriggerIfNonNil{
+				Ann: NewResKey(key.FuncDecl, key.RetNum, location)},
+			IsNamedReturn: ann.IsNamedReturn,
+			RetStmt:       ann.RetStmt,
+		},
+		Expr:         t.Expr,
+		Guards:       t.Guards.Copy(), // TODO: probably, we might not need a deep copy all the time
+		GuardMatched: t.GuardMatched,
+	}
+}
+
+// UseAsReturn is when a value flows to a point where it is returned from a function.
+// This consumer trigger can be used on top of two different sites: RetAnnotationKey &
+// ResAnnotationKey. RetAnnotationKey is the parameter site in the function declaration;
+// ResAnnotationKey is the argument site in the call expression. ResAnnotationKey is specifically
+// used for functions with contracts since we need to duplicate the sites for context sensitivity.
 type UseAsReturn struct {
 	TriggerIfNonNil
 	IsNamedReturn bool
@@ -487,12 +533,25 @@ func (u UseAsReturn) String() string {
 
 // Prestring returns this UseAsReturn as a Prestring
 func (u UseAsReturn) Prestring() Prestring {
-	retAnn := u.Ann.(RetAnnotationKey)
-	return UseAsReturnPrestring{
-		retAnn.FuncDecl.Name(),
-		retAnn.RetNum,
-		u.IsNamedReturn,
-		retAnn.FuncDecl.Type().(*types.Signature).Results().At(retAnn.RetNum).Name(),
+	switch key := u.Ann.(type) {
+	case RetAnnotationKey:
+		return UseAsReturnPrestring{
+			key.FuncDecl.Name(),
+			key.RetNum,
+			u.IsNamedReturn,
+			key.FuncDecl.Type().(*types.Signature).Results().At(key.RetNum).Name(),
+			"",
+		}
+	case ResAnnotationKey:
+		return UseAsReturnPrestring{
+			key.FuncDecl.Name(),
+			key.RetNum,
+			u.IsNamedReturn,
+			key.FuncDecl.Type().(*types.Signature).Results().At(key.RetNum).Name(),
+			key.Location.String(),
+		}
+	default:
+		panic(fmt.Sprintf("Expected RetAnnotationKey or ResAnnotationKey but see: %T", key))
 	}
 }
 
@@ -502,13 +561,22 @@ type UseAsReturnPrestring struct {
 	RetNum        int
 	IsNamedReturn bool
 	RetName       string
+	// Location is empty for a UseAsReturn enclosing RetAnnotationKey. Location points to the
+	// location of the result return at the call site for a UseAsReturn enclosing ResAnnotationKey.
+	Location string
 }
 
 func (u UseAsReturnPrestring) String() string {
+	var sb strings.Builder
+	sb.WriteString(fmt.Sprintf("returned from the function `%s`", u.FuncName))
 	if u.IsNamedReturn {
-		return fmt.Sprintf("returned from the function `%s` via the named return value `%s` in position %d", u.FuncName, u.RetName, u.RetNum)
+		sb.WriteString(fmt.Sprintf(" via the named return value `%s`", u.RetName))
 	}
-	return fmt.Sprintf("returned from the function `%s` in position %d", u.FuncName, u.RetNum)
+	sb.WriteString(fmt.Sprintf(" in position %d", u.RetNum))
+	if u.Location != "" {
+		sb.WriteString(fmt.Sprintf(" at %s", u.Location))
+	}
+	return sb.String()
 }
 
 // overriding position value to point to the raw return statement, which is the source of the potential error
--- a/annotation/consume_trigger.go
+++ b/annotation/consume_trigger.go
@@ -356,11 +356,12 @@ func (g GlobalVarAssignPrestring) String() string {
 	return fmt.Sprintf("assigned into the global variable `%s`", g.VarName)
 }
 
-// ArgPass is when a value flows to a point where it is passed as an argument to a function.
-// This consumer trigger can be used on top of two different sites: ParamAnnotationKey &
-// ArgAnnotationKey. ParamAnnotationKey is the parameter site in the function declaration;
-// ArgAnnotationKey is the argument site in the call expression. ArgAnnotationKey is specifically
-// used for functions with contracts since we need to duplicate the sites for context sensitivity.
+// ArgPass is when a value flows to a point where it is passed as an argument to a function. This
+// consumer trigger can be used on top of two different sites: ParamAnnotationKey &
+// CallSiteParamAnnotationKey. ParamAnnotationKey is the parameter site in the function
+// declaration; CallSiteParamAnnotationKey is the argument site in the call expression.
+// CallSiteParamAnnotationKey is specifically used for functions with contracts since we need to
+// duplicate the sites for context sensitivity.
 type ArgPass struct {
 	TriggerIfNonNil
 }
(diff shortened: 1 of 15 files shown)


assertion/function/analyzer.go
```diff
@@ -223,15 +225,29 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
...
 			errs = append(errs, r.err)
 		} else {
 			funcTriggers[r.index] = r.triggers
 			triggerCount += len(r.triggers)
+
+			funcObj, ok := pass.TypesInfo.ObjectOf(r.funcDecl.Name).(*types.Func)
+			if !ok {
+				continue
+			}
+			funcRes := r
```
lazaroclapp, line 240: Why the intermediate var here and not just: `funcResults[funcObj] = &r`?

assertion/function/analyzer.go
```diff
@@ -241,6 +257,175 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
...
+) {
+
+	// Find all the calls to contracted functions
+	// callsByCtrtFunc is a mapping contracted function -> caller -> call expression
+	callsByCtrtFunc := map[*types.Func]map[*types.Func]*ast.CallExpr{}
+	for funcObj, r := range funcResults {
+		for ctrFunc, calls := range findCallsToContractedFunctions(r.funcDecl, pass, funcContracts) {
+			for _, call := range calls {
+				// TODO: have to introduce v to make NilAway happy. Remove v when NilAway can
+				//  tolerate this.
```
lazaroclapp, line 281: Can we link an issue here explaining why `v` is needed? Or else add more detail to this comment on why this is needed now and why we think it might not be in the future?

assertion/function/analyzer.go
```diff
@@ -241,6 +257,175 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
...
+// but uses the new producers/consumers instead.
+func duplicateFullTriggersFromContractedFunctionsToCallers(
+	pass *analysis.Pass,
+	funcContracts functioncontracts.Map,
+	funcTriggers [][]annotation.FullTrigger,
+	funcResults map[*types.Func]*functionResult,
+) {
+
+	// Find all the calls to contracted functions
+	// callsByCtrtFunc is a mapping contracted function -> caller -> call expression
```
lazaroclapp, line 275: Just one question: Does this mean we support only one call-site per caller? What happens if a function calls a contracted function multiple times with different parameters? (That should probably be a test case too :) )

assertion/function/analyzer.go
```diff
@@ -241,6 +257,175 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
...
+			}
+		}
+	}
+
+	// For every contracted function, duplicate some of its full triggers (that involves param or
+	// return) into all the callers
+	dupTriggers := map[*types.Func][]annotation.FullTrigger{}
+	for ctrtFunc, calls := range callsByCtrtFunc {
+		// TODO: have to introduce r to make NilAway happy. Remove r when NilAway can tolerate
+		//  this.
```
lazaroclapp, line 297: Same as above.

assertion/function/analyzer.go
```diff
@@ -241,6 +257,175 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
...
+				// Store the duplicated full trigger
+				dupTriggers[caller] = append(dupTriggers[caller], dupTrigger)
+			}
+		}
+	}
+
+	// Update funcTriggers with duplicated triggers
+	for funcObj, triggers := range dupTriggers {
+		// TODO: have to introduce r to make NilAway happy. Remove r when NilAway can tolerate
+		//  this.
```
lazaroclapp, line 326: Probably worth linking an issue, then, rather than expanding and then repeating the explanation three times :)

assertion/function/analyzer.go
```diff
@@ -241,6 +257,175 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
...
+		r := funcResults[funcObj]
+		if r == nil {
+			continue
+		}
+		funcTriggers[r.index] = append(r.triggers, triggers...)
```
lazaroclapp, line 331: Wouldn't this make more sense as `funcTriggers[r.index] = append(funcTriggers[r.index], triggers...)`? Just in case we ever add another feature that appends to `funcTriggers[r.index]`? (I believe line 233 has already added `r.triggers` here, unless I am not understanding the control flow)

assertion/function/analyzer.go
```diff
@@ -241,6 +257,175 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
...
+		return true
+	})
+	return calls
+}
+
+// hasOnlyNonNilToNonNilContract returns whether the given function has only one contract that is
+// nonnil->nonnil.
+func hasOnlyNonNilToNonNilContract(funcContracts functioncontracts.Map, funcObj *types.Func) bool {
+	contracts, ok := funcContracts[funcObj]
+	if !ok || len(contracts) != 1 {
```
lazaroclapp, line 421: Do we need to restrict ourselves to there being no other contracts, or can we loop through them and just return `true` if any of the contracts present is `nonnil -> noninil`?

inference/engine.go
```diff
@@ -39,6 +39,10 @@ type Engine struct {
 	// converted to diagnostics and returned along with the current inferred map whenever users
 	// request it.
 	conflicts *conflictGrouping
+	// controlledTriggersBySite stores the set of controlled triggers for each site if the site is
+	// controls any triggers. This field is for internal use in the struct only and should not be
```
lazaroclapp, line 43: Can't parse the end of this first sentence: "if the site is controls any triggers". Should it be "if the site controls any triggers"?

annotation/consume_trigger.go
```diff
@@ -355,7 +356,11 @@ func (g GlobalVarAssignPrestring) String() string {
 	return fmt.Sprintf("assigned into the global variable `%s`", g.VarName)
 }
 
-// ArgPass is when a value flows to a point where it is passed as an argument to a function
+// ArgPass is when a value flows to a point where it is passed as an argument to a function.
+// This consumer trigger can be used on top of two different sites: ParamAnnotationKey &
+// ArgAnnotationKey. ParamAnnotationKey is the parameter site in the function declaration;
+// ArgAnnotationKey is the argument site in the call expression. ArgAnnotationKey is specifically
+// used for functions with contracts since we need to duplicate the sites for context sensitivity.
```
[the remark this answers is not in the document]
zzqatuber: Great point, renamed to `CallSiteParamAnnotationKey` and `CallSiteRetAnnotationKey`.

annotation/full_trigger.go
```diff
@@ -41,8 +52,12 @@ func (t *FullTrigger) Pos() token.Pos {
 }
 
 func (t *FullTrigger) String() string {
-	return fmt.Sprintf("[%T -> (%T@%d %s)]",
+	str := fmt.Sprintf("%T -> (%T@%d %s)",
 		t.Producer.Annotation, t.Consumer.Expr, t.Pos(), t.Consumer.Annotation.String())
+	if t.Controller != nil {
+		str += " if " + t.Controller.String() + " is nilable"
```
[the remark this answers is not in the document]
zzqatuber: 1. The `String()` of `FullTrigger` is totally for debugging.
2. Even the current representation is OK because the `Controller` is of `*CallSiteParamAnnotationKey` type and `String()` of `*CallSiteParamAnnotationKey` is already very compact.

annotation/key.go
```diff
@@ -62,6 +63,85 @@ func (k FieldAnnotationKey) String() string {
...
+
+// ParamName returns the *types.Var naming the parameter associate with this key
+// nilable(result 0)
+func (ak ArgAnnotationKey) ParamName() *types.Var {
+	return ak.FuncDecl.Type().(*types.Signature).Params().At(ak.ParamNum)
+}
+
+// Lookup looks this key up in the passed map, returning a Val
+func (ak ArgAnnotationKey) Lookup(annMap Map) (Val, bool) {
+	if paramVal, ok := annMap.CheckFuncParamAnn(ak.FuncDecl, ak.ParamNum); ok {
```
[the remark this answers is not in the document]
zzqatuber: Right, this line will be updated in the next diff in no-infer mode.

annotation/key.go
```diff
@@ -144,14 +224,51 @@ func (pk ParamAnnotationKey) MinimalString() string {
...
+// the return result for every call of the same function.
+type ResAnnotationKey struct {
+	FuncDecl *types.Func
+	RetNum   int // which result
+	Location token.Position
+}
+
+// Lookup looks this key up in the passed map, returning a Val
+func (rk ResAnnotationKey) Lookup(annMap Map) (Val, bool) {
+	if retVal, ok := annMap.CheckFuncRetAnn(rk.FuncDecl, rk.RetNum); ok {
```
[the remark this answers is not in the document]
zzqatuber: Same comment as for `ArgAnnotationKey.Lookup`

assertion/function/analyzer.go
```diff
@@ -86,6 +86,8 @@ type functionResult struct {
 	// declarations can be parallelized.
 	// TODO: remove this.
 	index int
+	// funcDecl is the function declaration in the package.
```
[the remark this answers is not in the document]
zzqatuber: You are correct.

assertion/function/analyzer.go
zzqatuber: Good question. Thanks to the special design choice of Go, `r` in each iteration is re-updated with a new value, which means the address of `r` never changes. This is to make sure we get the correct address in every iteration; otherwise if we do `&r`, we always have the same address, which does not change with the loop.

assertion/function/analyzer.go
zzqatuber: I agree.

assertion/function/analyzer.go
zzqatuber: I am not clear about what we should do with other contracts co-existing with `nonnil->nonnil`. In the future if we want to support multiple contracts, we should think of this problem more comprehensively and more likely we will not have this `hasOnlyNonNilToNonNilContract` function at all. Thus, it should be good to just have this function at the exact restriction level as we support at the current stage.

inference/engine.go
zzqatuber: Good catch.

testdata/src/go.uber.org/functioncontracts/inference/functioncontracts-with-inference.go
```diff
@@ -0,0 +1,125 @@
...
+// limitations under the License.
+
+// <nilaway contract enable>
+package inference
+
+import "math/rand"
+
+// Test the contracted function contains a full trigger nilable -> return 0.
+// contract(nonnil -> nonnil)
+func fooReturn(x *int) *int { // want "^ Annotation on Result 0.*\n.*Must be NILABLE.*\n.*AND.*\n.*Must be NONNIL.*NONNIL$"
```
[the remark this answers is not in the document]
zzqatuber: We don't have `a2` or `b2` in the error message, so the best thing we can do now is to directly put the line number, which, however, prevents future update of the test file.

Maybe we can have intermediate variable names when we have the new format of error messages? @sonalmahajan15

testdata/src/go.uber.org/functioncontracts/inference/functioncontracts-with-inference.go
```diff
@@ -0,0 +1,125 @@
...
+
+func barReturn2() {
+	var a2 *int
+	b2 := fooReturn(a2)
+	print(*b2) // "nilable value dereferenced" wanted
+}
+
+// Test the contracted function contains a full trigger param 0 -> nonnil.
+// contract(nonnil -> nonnil)
+func fooParam(x *int) *int { // want "^ Annotation on Param 0.*\n.*Must be NILABLE.*\n.*AND.*\n.*Must be NONNIL.*NONNIL$"
```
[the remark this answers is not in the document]
zzqatuber: Good point. I updated the test.

testdata/src/go.uber.org/functioncontracts/inference/functioncontracts-with-inference.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// Test the contracted function contains another contracted function.
+// contract(nonnil -> nonnil)
+func fooNested(x *int) *int {
+	return fooBase(x)
+}
+
+// contract(nonnil -> nonnil)
+func fooBase(x *int) *int { // want "^ Annotation on Result 0.*\n.*Must be NILABLE.*\n.*AND.*\n.*Must be NONNIL.*NONNIL$"
```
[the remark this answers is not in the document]
zzqatuber: updated.

assertion/function/analyzer.go
zzqatuber: No, it is totally OK to have multiple calls to a contracted functions with different parameters in a single function. Added a test case.

annotation/key.go
```diff
@@ -62,6 +63,85 @@ func (k FieldAnnotationKey) String() string {
...
+		ak.ParamNum, argname, ak.FuncDecl.Name(), ak.Location.String())
+}
+
+// MinimalString returns a string representation for this ArgAnnotationKey consisting only
+// of the word "arg" followed by the name of the parameter, if named, or its position otherwise
+func (ak ArgAnnotationKey) MinimalString() string {
+	if ak.ParamName() != nil && len(ak.ParamName().Name()) > 0 {
+		return fmt.Sprintf("arg `%s`", ak.ParamName().Name())
+	}
+	return fmt.Sprintf("arg %d", ak.ParamNum)
```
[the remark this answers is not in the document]
zzqatuber: Correct. We don't have such a test case. The only contract format that could make sense for such a function with unnamed parameter is for example (_ -> ...) but again this contract does not say anything useful on the parameter.

I copied the code basically from the existing `MinimalString()` method for `ParamAnnotationKey`. We can 
1) Keep as is, although the else branch would be not reached I believe.
2) Remove the else branch and the if condition, assuming we always have a named parameter for a `CallSiteAnnotationKey`.

Personally I am inclined to 1 because it requires no assumption and is thus decoupled with any further changes we could make in other places.

assertion/function/analyzer.go
zzqatuber: Enriched the code comment.

Ideally we should write this

```go
if _, ok := callsByCtrtFunc[ctrFunc]; !ok {
    callsByCtrtFunc[ctrFunc] = map[*types.Func]*ast.CallExpr{}
}
callsByCtrtFunc[ctrFunc][funcObj] = call // line 285
```

but NilAway would complain
```
Value read deeply from the variable `callsByCtrtFunc` lacking guarding at function/analyzer.go:285:5 (definitely nilable) and is written to at an index at function/analyzer.go:285:5 (must be nonnil)
```

assertion/function/analyzer.go
zzqatuber: The TODO is outdated. Actually NilAway's warning made sense so I added a nil check. Removed the TODO comment.

testdata/src/go.uber.org/functioncontracts/inference/functioncontracts-with-inference.go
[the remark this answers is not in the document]
sonalmahajan15: Yes, the variable names would likely be a part of the error message. I can see `a2` would definitely be, in the form of `unassigned variable a2`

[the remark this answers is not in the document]
lazaroclapp: Ok, let's make a small issue for a follow up fix here to include the variable name after Sonal's change lands. It will likely be a one line PR, but that's fine 🙂 (Issue can be a title and a link to this thread, no need to describe in detail)

annotation/key.go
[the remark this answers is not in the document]
lazaroclapp: Yes, I agree. Even if we go for 2, we would basically have to add some sort of assertion/check and panic if we ever see such malformed contract, whereas this is more robust.

We could even have a test cases of the style:

```
// Contract below isn't useful, since return is always nonnil and argument
// is ignored, but added to check we don't crash on unnamed parameters
// contract(nonnil -> nonnil)
func foo(_ *int) *int {
  return &1
}
```

assertion/function/analyzer.go
lazaroclapp: > Thanks to the special design choice of Go

A phrase that crops up all to often on NilAway development 😅 But, right, makes sense.

assertion/function/analyzer.go
lazaroclapp: @sonalmahajan15 This _seems_ like it should be handled by trackable expressions. Is it some weird trackable expression + guarding interaction? Do we have a JIRA ticket on this we might want to migrate to an OSS issue :)

assertion/function/analyzer.go
lazaroclapp: Fair enough. My intuition is that each `// contract` line will be interpreted independently in `AND` fashion, since that's what happens with multiple contracts in Java, but you are right this needs more thought and I am fine with the conservative code for now. Maybe then as a short flow up PR, let's have NilAway warn whenever any function has multiple contracts? Since otherwise they silently suppress one another.

assertion/function/analyzer.go
```diff
@@ -293,10 +300,10 @@ func duplicateFullTriggersFromContractedFunctionsToCallers(
 	// return) into all the callers
 	dupTriggers := map[*types.Func][]annotation.FullTrigger{}
 	for ctrtFunc, calls := range callsByCtrtFunc {
-		// TODO: have to introduce r to make NilAway happy. Remove r when NilAway can tolerate
-		//  this.
 		r := funcResults[ctrtFunc]
 		if r == nil {
+			// should not happen since funcResults should contain all the functions including any
+			// contracted functions.
```
lazaroclapp, line 306: If we are _certain_ this can't happen due to NilAway invariants... should we panic here instead? (The top level handler will convert that to a reported error rather than a crash, I believe and it has the advantage of making debugging easier if this assumption is ever broken).

assertion/function/analyzer.go
```diff
@@ -272,13 +272,20 @@ func duplicateFullTriggersFromContractedFunctionsToCallers(
 ) {
 
 	// Find all the calls to contracted functions
-	// callsByCtrtFunc is a mapping contracted function -> caller -> call expression
+	// callsByCtrtFunc is a mapping: contracted function -> caller -> all the call expressions
```
lazaroclapp, line 275: How does this work, though? Isn't `*ast.CallExpr{}` a single call expression rather than a list/set of expressions? Even the loop `for caller, callExpr := range calls` below gives one `callExpr` per `caller`, no?

testdata/src/go.uber.org/functioncontracts/inference/functioncontracts-with-inference.go
```diff
@@ -112,14 +107,45 @@ func barNested2() {
...
-	call(fooParamCalledInAnotherFunction(s))
+	var x *int
+	call(fooParamCalledInAnotherFunction(x))
 }
 
 func call(x *int) {}
+
+// Test a contracted function is called multiple times in another function.
+// contract(nonnil->nonnil)
+func fooReturnCalledMultipleTimesInTheSameFunction(x *int) *int { // want "^ Annotation on Result 0.*\n.*Must be NILABLE.*\n.*AND.*\n.*Must be NONNIL.*NONNIL$"
```
lazaroclapp, line 131: Can we get a version of these two functions where both callsites pass `nilable` and make sure two errors are reported? I am still not sure we are checking both calls within the same caller.

lazaroclapp (commented):
Seems good to me, but want to make sure we are supporting multiple callsites in the same caller.

testdata/src/go.uber.org/functioncontracts/inference/functioncontracts-with-inference.go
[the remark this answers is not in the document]
zzqatuber: See issue https://github.com/uber-go/nilaway/issues/35.

testdata/src/go.uber.org/functioncontracts/inference/functioncontracts-with-inference.go
zzqatuber: Nice point, included.

assertion/function/analyzer.go
zzqatuber: You are absolutely right! Nice catch! Fixed.

zzqatuber:
> Seems good to me, but want to make sure we are supporting multiple callsites in the same caller.

Nice catch @lazaroclapp! We did not, fixed it.

assertion/function/analyzer.go
zzqatuber: Included.

lazaroclapp (approved):
LGTM! 🎉

annotation/key.go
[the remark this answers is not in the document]
zzqatuber: Cool, added.

assertion/function/analyzer.go
sonalmahajan15: Ah yes, this is a known problem, where tracking non-literals is limited. We do have a JIRA ticket for this, which we should now migrate to OSS. Will do :)

PR #27, opened 2023-08-05
Store original AST expressions in assertion nodes

sonalmahajan15:
This PR adds original AST expressions in the assertion nodes to allow printing of positions of producer expressions.


lazaroclapp (approved):
LGTM. One high-level question: Within the analysis of a single target, are all ASTs persisted in memory by the nogo/open-source driver? Or are they usually garbage collected and keeping all these references in the assertion nodes is increasing memory usage? Also, do these persist intra-procedurally or only within each intra-procedural analysis?

(Basically, I am wondering why we used to go our of our way to not include the AST node as part of the assertion tree node before this change, and if there was a good reason for that, since it definitely seems simpler to have the AST node at hand :) )

assertion/function/assertiontree/assertion_node.go
```diff
@@ -73,6 +72,10 @@ type assertionNodeCommon struct {
 	parent          AssertionNode // this should be nil for the root
 	children        []AssertionNode
 	consumeTriggers []*annotation.ConsumeTrigger
+
+	// originalExpr stores the original call expression that prompted the creation of this assertion node
+	originalExpr ast.Expr // this should be nil for the root
```
yuxincs, line 77: Wonder if it's just better to call it `Expr`, since there are no "fake" expressions here anyways.

assertion/function/assertiontree/assertion_node.go
```diff
@@ -73,6 +72,10 @@ type assertionNodeCommon struct {
 	parent          AssertionNode // this should be nil for the root
 	children        []AssertionNode
 	consumeTriggers []*annotation.ConsumeTrigger
+
+	// originalExpr stores the original call expression that prompted the creation of this assertion node
```
yuxincs, line 76: "the original call expression", is this true? 🤔  I think, for example, for field assertion nodes this is a selector expression, right?

assertion/function/assertiontree/assertion_node.go
```diff
@@ -106,3 +109,7 @@ func (n *assertionNodeCommon) Size() int {
 	}
 	return size
 }
+
+func (n *assertionNodeCommon) BuildExpr() ast.Expr {
+	return n.originalExpr
+}
```
yuxincs, line 115: I do appreciate the encapsulations here, but I wonder if it's really necessary anymore since we no longer need polymorphisms here and I don't see such need for the near future.

We can just make `originalExpr` exported and use the field instead, which is a lot simpler. (we can always encapsulate later when there is a need)

yuxincs (approved):
Overall this looks great to me! 

I only have a few nits on further (debatable) simplifications.

yuxincs:
> Within the analysis of a single target, are all ASTs persisted in memory by the nogo/open-source driver?

Yes

nogo: https://github.com/bazelbuild/rules_go/blob/0da92cd6c14428b9838395e53e1a773fffbb37f0/go/tools/builders/nogo_main.go#L197-L203
open-source driver (i.e., golangci-lint): https://golangci-lint.run/usage/performance/#why-golangci-lint-is-so-fast

sonalmahajan15:
Closing this pull request since it was causing ambiguity in the error message regards to the position. Subsequently we also changed the error message format deeming this PR unnecessary.

PR #28, opened 2023-08-05
New error message format

sonalmahajan15:
This PR introduces a new, improved error message format with the aim of improving UX.

New message format:

```
Potential nil panic detected. Observed nil flow from source to dereference point:
	-> CONSUMER_POS: explanation text from producer and consumer
	-> CONSUMER_POS: explanation text from producer and consumer
        ...
	-> CONSUMER_POS: explanation text from producer and consumer
```

`conflict.go` is revamped to allow conformance to the new error message format. For simplicity of this PR, the grouping logic is removed. It will be added back in a follow up PR.

### DIFF
--- a/assertion/function/assertiontree/preprocess_blocks.go
+++ b/assertion/function/assertiontree/preprocess_blocks.go
@@ -85,7 +85,9 @@ func copyGraph(graph *cfg.CFG) *cfg.CFG {
 	// must have already been copied.
 	for i, newBlock := range newGraph.Blocks {
 		for _, succ := range graph.Blocks[i].Succs {
-			newBlock.Succs = append(newBlock.Succs, copiedBlocks[succ])
+			if c, ok := copiedBlocks[succ]; ok {
+				newBlock.Succs = append(newBlock.Succs, c)
+			}
 		}
 	}
 
--- a/assertion/function/assertiontree/preprocess_blocks.go
+++ b/assertion/function/assertiontree/preprocess_blocks.go
@@ -85,9 +85,7 @@ func copyGraph(graph *cfg.CFG) *cfg.CFG {
 	// must have already been copied.
 	for i, newBlock := range newGraph.Blocks {
 		for _, succ := range graph.Blocks[i].Succs {
-			if c, ok := copiedBlocks[succ]; ok {
-				newBlock.Succs = append(newBlock.Succs, c)
-			}
+			newBlock.Succs = append(newBlock.Succs, copiedBlocks[succ])
 		}
 	}
 
--- a/assertion/function/assertiontree/preprocess_blocks.go
+++ b/assertion/function/assertiontree/preprocess_blocks.go
@@ -549,7 +549,9 @@ func markSwitchStatements(graph *cfg.CFG, switchChildren map[ast.Node]*ast.Switc
 				},
 			}
 
-			caseBlockIdx = graph.Blocks[caseBlockIdx].Succs[1].Index
+			if blockSuccs := graph.Blocks[caseBlockIdx].Succs; blockSuccs != nil {
+				caseBlockIdx = blockSuccs[1].Index
+			}
 		}
 	}
 }
--- a/inference/conflict.go
+++ b/inference/conflict.go
@@ -98,7 +98,7 @@ func (n *nilFlow) String() string {
 
 func (c *conflict) String() string {
 	consumerPos := c.nilFlow.nonnilPath[len(c.nilFlow.nonnilPath)-1].position
-	producerPos := c.nilFlow.nilPath[len(c.nilFlow.nilPath)-1].position
+	producerPos := c.nilFlow.nilPath[0].position
 	return fmt.Sprintf(" Nonnil `%s` expected at \"%s\", but produced as nilable at \"%s\". Observed nil flow from "+
 		"source to dereference: %s", c.expr, consumerPos.String(), producerPos.String(), c.nilFlow.String())
 }
--- a/inference/conflict.go
+++ b/inference/conflict.go
@@ -17,6 +17,7 @@ package inference
 import (
 	"fmt"
 	"go/token"
+	"strings"
 
 	"go.uber.org/nilaway/annotation"
 	"go.uber.org/nilaway/util"
@@ -24,8 +25,8 @@ import (
 )
 
 type conflict struct {
-	position token.Pos // stores position where the error should be reported
-	nilFlow  nilFlow   // stores nil flow from source to dereference point
+	pos  token.Pos // stores position where the error should be reported (note that this field is used only within the current, and should NOT be exported)
+	flow nilFlow   // stores nil flow from source to dereference point
 }
 
 type nilFlow struct {
@@ -40,7 +41,7 @@ type node struct {
 }
 
 // newNode creates a new node object from the given producer and consumer Prestrings.
-// LocatedPrestring contains accurate information about the position and the reason why NilAway deemed that position
+// LocatedPrestring contains accurate information about the pos and the reason why NilAway deemed that pos
 // to be nilable. We use it if available, else we use the raw string representation available from the Prestring.
 func newNode(p annotation.Prestring, c annotation.Prestring) node {
 	nodeObj := node{}
@@ -78,7 +79,7 @@ func (n *node) String() string {
 		reasonStr = n.consumerRepr
 	}
 
-	return fmt.Sprintf("\n\t-> %s: %s", posStr, reasonStr)
+	return fmt.Sprintf("\t-> %s: %s", posStr, reasonStr)
 }
 
 // addNilPathNode adds a new node to the nil path.
@@ -88,6 +89,7 @@ func (n *nilFlow) addNilPathNode(p annotation.Prestring, c annotation.Prestring)
 	// Note that in the implication graph, we traverse backwards from the point of conflict to the source of nilability.
 	// Therefore, they are added in reverse order from what the program flow would look like. To account for this we
 	// prepend the new node to nilPath because we want to print the program flow in its correct (forward) order.
+	// TODO: instead of prepending here, we can reverse the nilPath slice while printing.
 	n.nilPath = append([]node{nodeObj}, n.nilPath...)
 }
 
@@ -97,21 +99,21 @@ func (n *nilFlow) addNonNilPathNode(p annotation.Prestring, c annotation.Prestri
 	n.nonnilPath = append(n.nonnilPath, nodeObj)
 }
 
-// String converts a nilFlow to a string representation, where each entry is the flow of the form: `<pos>: <reason>`
+// String converts a flow to a string representation, where each entry is the flow of the form: `<pos>: <reason>`
 func (n *nilFlow) String() string {
-	flow := ""
+	var flow []string
 	for _, nodes := range [...][]node{n.nilPath, n.nonnilPath} {
 		for _, nodeObj := range nodes {
-			flow += nodeObj.String()
+			flow = append(flow, nodeObj.String())
 		}
 	}
-	return flow
+	return strings.Join(flow, "\n")
 }
 
 func (c *conflict) String() string {
-	consumerPos := c.nilFlow.nonnilPath[len(c.nilFlow.nonnilPath)-1].position
+	consumerPos := c.flow.nonnilPath[len(c.flow.nonnilPath)-1].position
 	return fmt.Sprintf(" Potential nil panic at \"%s\". Observed nil flow from "+
-		"source to dereference: %s", consumerPos.String(), c.nilFlow.String())
+		"source to dereference: %s", consumerPos.String(), c.flow.String())
 }
 
 type conflictList struct {
@@ -121,11 +123,11 @@ type conflictList struct {
 func (l *conflictList) addSingleAssertionConflict(pass *analysis.Pass, trigger annotation.FullTrigger) {
 	t := fullTriggerAsPrimitive(pass, trigger)
 	c := conflict{
-		position: t.Pos,
-		nilFlow:  nilFlow{},
+		pos:  t.Pos,
+		flow: nilFlow{},
 	}
 
-	c.nilFlow.addNonNilPathNode(t.ProducerRepr, t.ConsumerRepr)
+	c.flow.addNonNilPathNode(t.ProducerRepr, t.ConsumerRepr)
 
 	l.conflicts = append(l.conflicts, c)
 }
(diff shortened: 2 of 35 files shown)


assertion/function/assertiontree/preprocess_blocks.go
```diff
@@ -85,7 +85,9 @@ func copyGraph(graph *cfg.CFG) *cfg.CFG {
 	// must have already been copied.
 	for i, newBlock := range newGraph.Blocks {
 		for _, succ := range graph.Blocks[i].Succs {
-			newBlock.Succs = append(newBlock.Succs, copiedBlocks[succ])
+			if c, ok := copiedBlocks[succ]; ok {
+				newBlock.Succs = append(newBlock.Succs, c)
+			}
```
sonalmahajan15, line 90: Fix applied for a NilAway reported error, which seems to be a valid one. Not sure why it did not get reported earlier, though.

inference/engine.go
```diff
@@ -406,11 +406,9 @@ func GobRegister() {
 	gob.RegisterName(nextStr(), FalseBecauseShallowConstraint{})
 	gob.RegisterName(nextStr(), FalseBecauseDeepConstraint{})
 	gob.RegisterName(nextStr(), FalseBecauseAnnotation{})
-	gob.RegisterName(nextStr(), FalseBecauseMyopia{})
```
sonalmahajan15, line 409: The structs `TrueBecauseMyopia` and `FalseBecauseMyopia` and their methods are removed because they were unused.

inference/conflict.go
```diff
@@ -19,148 +19,177 @@ import (
...
 }
 
-// addConflict adds a conflict to a conflict group identified by the conflict's key.
-func (t *conflictGrouping) addConflict(conflict conflict) {
-	key := conflict.key()
-	t.groupMap[key] = append(t.groupMap[key], conflict)
+func (n *node) String() string {
+	posStr := "<no pos info>"
+	reasonStr := "<no reason info>"
+	if n.position.IsValid() {
```
lazaroclapp, line 56: Is this one of those cases where a method supports nilable receiver? (since I see from the function above that `n` could just be `node{}`)

inference/conflict.go
```diff
@@ -19,148 +19,177 @@ import (
...
-	for _, conflicts := range t.groupMap {
-		site := conflicts[0].getSiteMessage()
-		producer := conflicts[0].getProducerMessage()
-		consumer := ""
-		for _, c := range conflicts {
-			consumer += c.getConsumerMessage()
-		}
+// addNilPathNode adds a new node to the nil path.
+// LocatedPrestring contains accurate information about the position and the reason why NilAway deemed that position
+// to be nilable. We use it if available, else we use the raw string representation available from the Prestring `p`.
```
lazaroclapp, line 67: Feels like this doc should be in the `newNode` function now. No?

inference/conflict.go
```diff
@@ -19,148 +19,177 @@ import (
...
+func (n *nilFlow) addNilPathNode(p annotation.Prestring) {
+	nodeObj := newNode(p)
 
-		diagnostics = append(diagnostics, analysis.Diagnostic{
-			Pos:     conflicts[0].getPosition(),
-			Message: fmt.Sprintf("%s%s%s", site, producer, consumer),
-		})
-	}
-	return diagnostics
+	// Note that we prepend the new node to nilPath because we want to print the flow in the order in which it was discovered.
```
lazaroclapp, line 71: This is missing some info, which I think is that these nodes get added starting from the conflict node, so they are added in reverse order from what the program flow would look like.

inference/conflict.go
```diff
@@ -19,148 +19,177 @@ import (
...
+	// must take this into consideration while printing the flow.)
+	var queue []ExplainedBool
+	queue = append(queue, nilExplanation)
+
+	for len(queue) > 0 {
+		e := queue[0]
+		queue = queue[1:]
+
+		t := e.getPrimitiveFullTrigger()
+		if t.ConsumerRepr != nil && t.ProducerRepr != nil {
```
lazaroclapp, line 138: Can I get a one sentence reminder in comments here about what these two cases for `t` represent? :)

inference/explained_bool.go
```diff
@@ -113,44 +140,48 @@ func (f FalseBecauseDeepConstraint) String() string {
...
-	return fmt.Sprintf("NILABLE because myopic annotation inference for package %s decided so", t.PkgName)
-}
-
-// FalseBecauseMyopia is used as the label for a site X that a myopic inference procedure decided should
-// be nonnil
-type FalseBecauseMyopia struct {
-	ExplainedFalse
-	PkgName string
+func (f FalseBecauseAnnotation) getPrimitiveFullTrigger() primitiveFullTrigger {
+	return primitiveFullTrigger{Pos: f.Pos}
```
lazaroclapp, line 182: Weird that this line isn't reached by tests, actually (not something to fix in this PR but it strikes me as curious. When is `FalseBecauseAnnotation` used?)

testdata/src/go.uber.org/anonymousfunction/anonymous_functions_assignment.go
```diff
@@ -18,15 +18,15 @@ package anonymousfunction
 
 func testSimpleAssignment() {
 	var t1 *int
-	a := func() { // want "value is read from a variable that was never assigned to"
-		print(*t1)
+	a := func() {
+		print(*t1) //want "read from a variable that was never assigned to"
```
lazaroclapp, line 22: Interesting. What is causing us to switch these to get reported here vs at lambda declaration? (This is better, but I wonder if it's an intentional change or some consequence of the new error reporting method)

Edit: Just realized that the code above keeps updating the conflict position for as long as it finds nodes in its non-nil path, so the reporting position of the conflict ends up being that of the final non-nil consumer in the flow, which is what we want. Nice! (Maybe worth a once sentence comment in the code where this happens)

lazaroclapp (commented):
This is fantastic 🚀 ! Looks good to me, but I had a few questions / requests for more comments, so I'll go ahead and approve it once those are answered.

Btw, I looked at the actual code in detail, but the changes to testdata are a lot, so I mostly just eyeballed them. If you think there might be issues there, I can go line by line 😅

inference/conflict.go
sonalmahajan15: Not really. `newNode` is returning a struct, not a pointer, so the receiver `n` here of `String()` is never nil. Also, `n.position`, which is of type `token.Position`, is also a struct, not a pointer, so we can safely invoke the method `IsValid()` on it.

inference/conflict.go
sonalmahajan15: Ah, good catch. Moved it to `newNode`.

inference/conflict.go
sonalmahajan15: Your are right. Modified the comment to explain this better.

inference/conflict.go
sonalmahajan15: Yes, added a comment :)

inference/explained_bool.go
sonalmahajan15: Hmm.. `FalseBecauseAnnotation` is used when a program site is explicitly annotated as `nonnil`. I can see a couple of instances in our test cases, where this is present. So, not sure why isn't it covered by tests. I'll look into it.

inference/conflict.go
lazaroclapp: Ah, right, too used to Java reference types. I did mean `n.position` would be `nil` by default, but, you are right, since it is a struct type and not a pointer to struct, it will just be an empty struct itself...

testdata/src/go.uber.org/anonymousfunction/anonymous_functions_assignment.go
sonalmahajan15: Exactly, you got the reasoning right. Added a comment explaining this idea.

sonalmahajan15:
> This is fantastic 🚀 ! Looks good to me, but I had a few questions / requests for more comments, so I'll go ahead and approve it once those are answered.
> 
> Btw, I looked at the actual code in detail, but the changes to testdata are a lot, so I mostly just eyeballed them. If you think there might be issues there, I can go line by line 😅

Thanks @lazaroclapp. I have addressed / answered your comments. Please feel free to do a second pass.

I know, the test cases do have a lot of changes -- they are either about moving the position of where the error gets reported or modifying the text of `want` messages to suit the new format. I have ensured that there is nothing unexpected and the test cases pass as well, so feel free to not inspect them closely. :)

lazaroclapp (approved):
LGTM! (Though I still have my questions about that coverage issue for the `FalseBecauseAnnotation` explained bool...)

sonalmahajan15:
> LGTM! (Though I still have my questions about that coverage issue for the `FalseBecauseAnnotation` explained bool...)

I checked this issue and turns out that we did not have a viable test case for covering `FalseBecauseAnnotation` in our test suite, particularly in the full infer mode. I have added a test case in `testdata/src/go.uber.org/inference/inference.go` for the same, which now shows complete coverage of `FalseBecauseAnnotation`.

inference/conflict.go
```diff
@@ -19,148 +19,189 @@ import (
...
-// groupKey stores the type of the key representing a conflict group.
-// (Note that the type is currently set to any since the producer types of singleAssertionConflict and overconstrainedConflict
-// are not the same. Consolidating the two types of conflicts will help solve this problem (tracked in .)
-type groupKey any
-
-// conflictGrouping stores groups of conflicts represented by their keys in a map. It also provides methods to add conflicts
-// to the groups and to generate diagnostics from the groups.
-type conflictGrouping struct {
-	groupMap map[groupKey][]conflict
+type conflict struct {
```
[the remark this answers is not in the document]
sonalmahajan15: I like this. I'll create a refactoring diff at the top of this PR stack.

inference/explained_bool.go
```diff
@@ -29,6 +30,8 @@ import (
 type ExplainedBool interface {
 	Val() bool
 	String() string
+	getPrimitiveFullTrigger() primitiveFullTrigger
+	getExplainedBool() ExplainedBool
```
yuxincs, line 34:
```suggestion
	DeeperReason() ExplainedBool
```
Right? because `ExplainedBool.getExplainedBool` reads a little weird to me ;)

util/util.go
```diff
@@ -451,9 +453,27 @@ func IsLiteral(expr ast.Expr, literals ...string) bool {
...
+	return position
+}
+
 var codeReferencePattern = regexp.MustCompile("\\`(.*?)\\`")
 var pathPattern = regexp.MustCompile(`"(.*?)"`)
-var nilabilityPattern = regexp.MustCompile(`([\(|^\t](?i)(definitely\s|must\sbe\s)(nilable|nonnil)[\)]?)`)
+var nilabilityPattern = regexp.MustCompile(`([\(|^\t](?i)(found\s|must\sbe\s)(nilable|nonnil)[\)]?)`)
 
 // PrettyPrintErrorMessage is used in error reporting to post process and pretty print the output with colors
 func PrettyPrintErrorMessage(msg string) string {
```
yuxincs, line 479: Is it possible that we integrate (in the future of course) this pretty print logic into the diagnostics engine organically? 

Post-process via regex always feels a bit hacky to me :)

inference/conflict.go
```diff
@@ -19,148 +19,190 @@ import (
...
+	position token.Pos // stores position where the error should be reported
```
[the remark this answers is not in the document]
sonalmahajan15: Fair point. Renamed and added comment.

inference/conflict.go
```diff
@@ -19,148 +19,190 @@ import (
...
+	nilFlow  nilFlow   // stores nil flow from source to dereference point
```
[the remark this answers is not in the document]
sonalmahajan15: Done

util/util.go
sonalmahajan15: Agreed, should be doable. I'll try and incorporate this in the planned refactor.

inference/conflict.go
```diff
@@ -19,148 +19,190 @@ import (
...
+		} else {
+			c.nilFlow.addNilPathNode(annotation.LocatedPrestring{
+				Contained: e,
+				Location:  util.TruncatePosition(pass.Fset.Position(t.Pos)),
+			}, nil)
+		}
+
+		if e.getExplainedBool() != nil {
+			queue = append(queue, e.getExplainedBool())
+		}
```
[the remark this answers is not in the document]
sonalmahajan15: Agreed. Done

inference/explained_bool.go
sonalmahajan15: Good suggestion. I just un-exported `deeperReason()` since it is only used locally from within the pakcage.

inference/conflict.go
```diff
@@ -19,148 +19,190 @@ import (
...
+		reasonStr = n.producerRepr
+	} else if len(n.consumerRepr) > 0 {
+		reasonStr = n.consumerRepr
 	}
-}
 
-// getSiteMessage returns the message for the site of the conflict. For a singleAssertionConflict, this is an empty string.
-func (t *singleAssertionConflict) getSiteMessage() string {
-	return ""
+	return fmt.Sprintf("\n\t-> %s: %s", posStr, reasonStr)
```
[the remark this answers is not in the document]
sonalmahajan15: That's a great suggestion. Addressed :)

inference/conflict.go
```diff
@@ -19,148 +19,190 @@ import (
...
-func (t *singleAssertionConflict) getPosition() token.Pos {
-	return t.trigger.Pos
+// addNilPathNode adds a new node to the nil path.
+func (n *nilFlow) addNilPathNode(p annotation.Prestring, c annotation.Prestring) {
+	nodeObj := newNode(p, c)
+
+	// Note that in the implication graph, we traverse backwards from the point of conflict to the source of nilability.
+	// Therefore, they are added in reverse order from what the program flow would look like. To account for this we
+	// prepend the new node to nilPath because we want to print the program flow in its correct (forward) order.
+	n.nilPath = append([]node{nodeObj}, n.nilPath...)
```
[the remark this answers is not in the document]
sonalmahajan15: > But I think this does not add too much overhead anyway.
That is true, it is quite inexpensive, but might as well do it the right way. Added a TODO to act upon later.

inference/conflict.go
```diff
@@ -19,148 +19,190 @@ import (
...
+	return flow
 }
 
-// key returns the key of the conflict group that this conflict belongs to.
-func (t *singleAssertionConflict) key() groupKey {
-	return t.originalTrigger.Producer
+func (c *conflict) String() string {
+	consumerPos := c.nilFlow.nonnilPath[len(c.nilFlow.nonnilPath)-1].position
+	return fmt.Sprintf(" Potential nil panic at \"%s\". Observed nil flow from "+
+		"source to dereference: %s", consumerPos.String(), c.nilFlow.String())
```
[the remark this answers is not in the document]
sonalmahajan15: True. I'll prefer doing a refactoring at the top of the stack, or as a follow-up after this existing stack lands.

PR #30, opened 2023-08-05
Grouping for error messages

sonalmahajan15:
This PR adds back the grouping for error messages after adapting the logic to the new format.

[Depends on PR #28]

### DIFF
--- a/inference/conflict.go
+++ b/inference/conflict.go
@@ -41,7 +41,7 @@ type node struct {
 }
 
 // newNode creates a new node object from the given producer and consumer Prestrings.
-// LocatedPrestring contains accurate information about the position and the reason why NilAway deemed that position
+// LocatedPrestring contains accurate information about the pos and the reason why NilAway deemed that pos
 // to be nilable. We use it if available, else we use the raw string representation available from the Prestring.
 func newNode(p annotation.Prestring, c annotation.Prestring) node {
 	nodeObj := node{}
--- a/inference/conflict.go
+++ b/inference/conflict.go
@@ -27,7 +27,7 @@ import (
 type conflict struct {
 	pos  token.Pos // stores position where the error should be reported (note that this field is used only within the current, and should NOT be exported)
 	flow nilFlow   // stores nil flow from source to dereference point
-	similarConflicts []conflict // stores other conflicts that are similar to this one
+	similarConflicts []*conflict // stores other conflicts that are similar to this one
 }
 
 type nilFlow struct {
@@ -126,20 +126,19 @@ func (c *conflict) String() string {
 	// build string for similar conflicts (i.e., conflicts with the same nil path)
 	similarConflictsString := ""
 	if len(c.similarConflicts) > 0 {
-		similarPos := ""
-		for _, s := range c.similarConflicts {
-			similarPos += fmt.Sprintf("\"%s\", ", s.flow.nonnilPath[len(s.nilFlow.nonnilPath)-1].consumerPosition.String())
-		}
-		// remove trailing comma and space
-		similarPos = strings.TrimSuffix(similarPos, ", ")
-
-		// replace last comma with "and"
-		lastComma := strings.LastIndex(similarPos, ",")
-		if lastComma != -1 {
-			similarPos = similarPos[:lastComma] + " and" + similarPos[lastComma+1:]
+		similarPos := make([]string, len(c.similarConflicts))
+		for i, s := range c.similarConflicts {
+			similarPos[i] = fmt.Sprintf("\"%s\"", s.flow.nonnilPath[len(s.flow.nonnilPath)-1].consumerPosition.String())
 		}
 
-		similarConflictsString = fmt.Sprintf("\n\n(Same nil source could also cause potential nil panic(s) at %d other place(s): %s.)", len(c.similarConflicts), similarPos)
+		posString := strings.Join(similarPos[:len(similarPos)-1], ", ")
+		if len(similarPos) > 1 {
+			posString = posString + ", and "
+		}
+		posString = posString + similarPos[len(similarPos)-1]
+
+		similarConflictsString = fmt.Sprintf("\n\n(Same nil source could also cause potential nil panic(s) at %d "+
+			"other place(s): %s.)", len(c.similarConflicts), posString)
 	}
 
 	return fmt.Sprintf(" Potential nil panic at \"%s\". Observed nil flow from "+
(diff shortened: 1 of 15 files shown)


inference/conflict.go
```diff
@@ -87,20 +97,38 @@ func (n *nilFlow) String() string {
...
+
+	// build string for similar conflicts (i.e., conflicts with the same nil path)
+	similarConflictsString := ""
+	if len(c.similarConflicts) > 0 {
+		similarPos := ""
+		for _, s := range c.similarConflicts {
+			similarPos += fmt.Sprintf("\"%s\", ", s.nilFlow.nonnilPath[len(s.nilFlow.nonnilPath)-1].position.String())
+		}
+		// remove trailing comma and space
+		similarPos = strings.TrimSuffix(similarPos, ", ")
```
lazaroclapp, line 115: Does Go not have a string join function or similar? It feels kind of hacky to build the string with commas and then remove/replace stuff, maybe even adding an index to the for loop and then appending either `", "` for `i < len - 2`, " and " for `i == len - 2` and nothing for `i == len - 1` would be better?

inference/conflict.go
```diff
@@ -185,11 +213,38 @@ func (l *conflictList) addOverconstraintConflict(nilExplanation ExplainedBool, n
...
+			existingConflict.addSimilarConflict(c)
+			conflictsMap[s] = existingConflict
+		} else {
+			conflictsMap[s] = c
+		}
+	}
+
+	// update groupedConflicts list with grouped groupedConflicts
+	var groupedConflicts []conflict
+	for _, c := range conflictsMap {
```
lazaroclapp, line 246: Use of map iteration here will lead to non-deterministic printing of errors. Can we iterate over `allConflicts` instead, skipping those not in `conflictsMap`? Or could we even populate this list in the for loop above on the else branch? Might need switching some stuff to pointer types :)

inference/conflict.go
```diff
@@ -17,16 +17,18 @@ package inference
...
 )
 
 type conflict struct {
-	position token.Pos // stores position where the error should be reported
-	expr     string    // stores expression that is overcontrained (i.e., expected to be nonnil, but found nilable)
-	nilFlow  nilFlow   // stores nil flow from source to dereference point
+	position         token.Pos  // stores position where the error should be reported
+	expr             string     // stores expression that is overcontrained (i.e., expected to be nonnil, but found nilable)
+	nilFlow          nilFlow    // stores nil flow from source to dereference point
+	similarConflicts []conflict // stores other conflicts that are similar to this one
```
lazaroclapp, line 31: Are we duplicating a lot of data by having this be an array of structs rather than pointers to structs? What's the Go idiomatic approach here?

testdata/src/go.uber.org/structinit/local/local.go
```diff
@@ -28,12 +28,15 @@ type A struct {
 
 func m() {
 	var b = &A{}
+	// TODO: all errors because of "aptr" unassigned at struct initialization are grouped and reported on the line below,
+	// which is not correct. This should be fixed after https://github.com/uber-go/nilaway/issues/29 is implemented,
+	// and struct initi producer expressions are updated accordingly with the original AST expressions.
```
lazaroclapp, line 33:
```suggestion
	// and struct init producer expressions are updated accordingly with the original AST expressions.
```

testdata/src/go.uber.org/structinit/local/local.go
```diff
@@ -28,12 +28,15 @@ type A struct {
...
 	print(b.aptr.ptr) //want "unassigned at struct initialization"
```
lazaroclapp, line 34: Let's add to the want regexp here a check for `"similar nil problem(s) at 6 other place(s)"`, just to make sure none of these errors go away (we can also do it for the tests above, but I am more worried about this one)

inference/conflict.go
sonalmahajan15: Actually there is. I was not aware of it. Updated the above code to now use `strings.Join()`

inference/conflict.go
sonalmahajan15: Good point. From what I know, there is not one governing Go idiomatic approach. It depends on the particular case. But I think for ours, we can definitely use a slice of struct pointers to save space and modify the object directly. Updated.

inference/conflict.go
sonalmahajan15: Good point. Done :)

testdata/src/go.uber.org/structinit/local/local.go
sonalmahajan15: Done

inference/conflict.go
```diff
@@ -198,11 +232,51 @@ func (l *conflictList) addOverconstraintConflict(nilExplanation ExplainedBool, n
...
 	return diagnostics
 }
+
+// groupConflicts groups conflicts with the same nil path together and update conflicts list.
+func groupConflicts(allConflicts []conflict) []conflict {
+	conflictsMap := make(map[string]int)  // key: nil path string, value: index in `allConflicts`
+	indicesToIgnore := make(map[int]bool) // indices of conflicts to be ignored from `allConflicts`, since they are grouped with other conflicts
+
+	for i, c := range allConflicts {
+		key := pathString(c.nilFlow.nilPath)
```
yuxincs, line 255: Can we move the utility function `pathString` closer to this?

testdata/src/go.uber.org/structinit/local/local.go
yuxincs: +1

inference/conflict.go
```diff
@@ -198,11 +232,51 @@ func (l *conflictList) addOverconstraintConflict(nilExplanation ExplainedBool, n
...
+		if existingConflictIndex, ok := conflictsMap[key]; ok {
+			// Grouping condition satisfied. Add new conflict to `similarConflicts` in `existingConflict`, and update groupedConflicts map
+			allConflicts[existingConflictIndex].addSimilarConflict(c)
+			indicesToIgnore[i] = true
+		} else {
+			conflictsMap[key] = i
+		}
+	}
+
+	// update groupedConflicts list with grouped groupedConflicts                                                                                            String         ts
```
yuxincs, line 274:
```suggestion
	// update groupedConflicts list with grouped groupedConflicts                                                                                           
```

typo?

yuxincs (approved):
LGTM except one nit and a request to add group information to the want strings in order not to miss any errors

inference/conflict.go
sonalmahajan15: Yes! Not sure how it get introduced, though

inference/conflict.go
sonalmahajan15: Done

testdata/src/go.uber.org/structinit/local/local.go
sonalmahajan15: Done

PR #31, opened 2023-08-05
Unify error reporting in both infer and no-infer mode

sonalmahajan15:
This diff unifies the error reporting in infer and no-infer modes to make the architecture easier for maintenance. This involves a lot of changes to the `want` messages in the unit tests.

Example of how the error message for the no-infer case would look like now:
```
type bar struct {
        ...
	myr  *bar
}


// nilable(a)
func foo(a *bar) {
	myBar := &bar{}
        ...
	myBar.myr = a // "assigned into the field `myr`"
        ...
}
```

```
Potential nil panic at "annotationparse/annotationparse.go:49:14". Observed nil flow from source to dereference:
        	-> annotationparse/annotationparse.go:49:14: read from function parameter `a`, assigned into the field `myr`
```

[Depends on PR #30]

### DIFF
--- a/accumulation/analyzer.go
+++ b/accumulation/analyzer.go
@@ -170,6 +170,8 @@ func errorsToDiagnostics(errs []error) []analysis.Diagnostic {
 // buildDiagnostics takes a list of FullTriggers, which are assumed to already have been checked
 // and are known to fail, and returns a slice of the appropriate diagnostics for all the triggers.
 func buildDiagnostics(pass *analysis.Pass, triggers []annotation.FullTrigger) []analysis.Diagnostic {
+	// setting NoGrouping to true since the no-infer mode is used via unit tests currently, and we actually want
+	// separate errors for better testability
 	conflicts := inference.ConflictList{NoGrouping: true}
 	for _, trigger := range triggers {
 		conflicts.AddSingleAssertionConflict(pass, trigger)
--- a/accumulation/analyzer.go
+++ b/accumulation/analyzer.go
@@ -170,14 +170,11 @@ func errorsToDiagnostics(errs []error) []analysis.Diagnostic {
 // buildDiagnostics takes a list of FullTriggers, which are assumed to already have been checked
 // and are known to fail, and returns a slice of the appropriate diagnostics for all the triggers.
 func buildDiagnostics(pass *analysis.Pass, triggers []annotation.FullTrigger) []analysis.Diagnostic {
-	var errors []analysis.Diagnostic
+	conflicts := inference.ConflictList{NoGrouping: true}
 	for _, trigger := range triggers {
-		errors = append(errors, analysis.Diagnostic{
-			Pos:     trigger.Consumer.Pos(),
-			Message: trigger.BuildStringRepr(pass),
-		})
+		conflicts.AddSingleAssertionConflict(pass, trigger)
 	}
-	return errors
+	return conflicts.Diagnostics()
 }
 
 // checkErrors iterates over a set of full triggers, checking each one against a given annotation
--- a/annotation/full_trigger.go
+++ b/annotation/full_trigger.go
@@ -95,22 +95,6 @@ func (t *FullTrigger) truncatedProducerPos(pass *analysis.Pass) token.Position {
 	return util.PosToLocation(t.Producer.Expr.Pos(), pass)
 }
 
-// BuildStringRepr returns a string representation of a `FullTrigger` given an `analysis.Pass` to help Lookup its objects
-func (t *FullTrigger) BuildStringRepr(pass *analysis.Pass) string {
-	if t.Producer.Expr == nil {
-		return fmt.Sprintf("Possible nil flow to %s: nilable value %s",
-			t.truncatedConsumerPos(pass),
-			t.Consumer.Annotation.String())
-	}
-	// TODO: switch the order here to "value <produced> and <consumed>" instead of "nilable value <consumed>, earlier <produced>"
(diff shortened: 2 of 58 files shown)


inference/conflict.go
```diff
@@ -149,7 +152,8 @@ func (l *conflictList) addSingleAssertionConflict(pass *analysis.Pass, trigger a
 	l.conflicts = append(l.conflicts, c)
 }
 
-func (l *conflictList) addOverconstraintConflict(nilExplanation ExplainedBool, nonnilExplanation ExplainedBool, pass *analysis.Pass) {
+// AddOverconstraintConflict adds a new overconstraint conflict to the list of conflicts.
+func (l *ConflictList) AddOverconstraintConflict(nilExplanation ExplainedBool, nonnilExplanation ExplainedBool, pass *analysis.Pass) {
```
lazaroclapp, line 156: Nothing currently calls this from outside this package. This is just done for consistency with `AddOverconstraintConflict` being visible. Correct? (Good by me, just want to make sure)

testdata/src/go.uber.org/annotationparse/annotationparse.go
```diff
@@ -46,7 +46,7 @@ func foo(a, b *bar, c *bar, d, e *bar) (f, g *bar, h *bar) {
 	myBar.jar = a
 	myBar.karp = a
 	myBar.lug = a
-	myBar.myr = a //want "nilable value assigned into the field `myr`"
+	myBar.myr = a //want "assigned into the field `myr`"
```
lazaroclapp, line 49: I am curious, how do the full error messages look like here on no-infer mode? Presumably they will be a two locations flow (from directly NILABLE to directly NONNIL, right?). Could we get an example into the PR description?

lazaroclapp (commented):
Why did code coverage go down by >1%? There isn't a lot of new code added in this PR. Are we maybe missing some code to delete which is now unreachable?

inference/conflict.go
sonalmahajan15: That's right. :)

testdata/src/go.uber.org/annotationparse/annotationparse.go
sonalmahajan15: Now with the producer + consumer concatenation, it will be a single line. Added an example in the PR description.

accumulation/analyzer.go
```diff
@@ -170,14 +170,11 @@ func errorsToDiagnostics(errs []error) []analysis.Diagnostic {
 // buildDiagnostics takes a list of FullTriggers, which are assumed to already have been checked
 // and are known to fail, and returns a slice of the appropriate diagnostics for all the triggers.
 func buildDiagnostics(pass *analysis.Pass, triggers []annotation.FullTrigger) []analysis.Diagnostic {
-	var errors []analysis.Diagnostic
+	conflicts := inference.ConflictList{NoGrouping: true}
```
yuxincs, line 173: Right, as we discussed, separating the diagnostics engine into its own package would make the separation a lot better!

yuxincs: We don't want grouping since no-infer mode is mostly used for tests and we actually want separate errors for better testability, right? If so, can we add a comment here for future readers?

annotation/full_trigger.go
```diff
@@ -95,22 +95,6 @@ func (t *FullTrigger) truncatedProducerPos(pass *analysis.Pass) token.Position {
...
-	}
-	// TODO: switch the order here to "value <produced> and <consumed>" instead of "nilable value <consumed>, earlier <produced>"
-	// only delaying because all of the tests' expectations will have to be modified
-	return fmt.Sprintf("Possible nil flow from %s to %s: nilable value %s, earlier %s",
-		t.truncatedProducerPos(pass),
-		t.truncatedConsumerPos(pass),
-		t.Consumer.Annotation.String(),
-		t.Producer.Annotation.String())
-}
-
```
yuxincs, line 113: Love this unification / simplification 👍

inference/conflict.go
yuxincs: Plus once we move the diagnostic processing logic to a separate package we'll need this exported

inference/conflict.go
```diff
@@ -230,14 +234,20 @@ func (l *conflictList) addOverconstraintConflict(nilExplanation ExplainedBool, n
...
 
-	// group conflicts with the same nil path together for concise reporting
-	groupedConflicts := groupConflicts(l.conflicts)
+	var conflicts []conflict
+	if l.NoGrouping {
+		conflicts = l.conflicts
+	} else {
+		// group conflicts with the same nil path together for concise reporting
+		conflicts = groupConflicts(l.conflicts)
+	}
```
yuxincs, line 247:
```suggestion
	conflicts := l.conflicts
	if !l.NoGrouping {
		// group conflicts with the same nil path together for concise reporting
		conflicts = groupConflicts(l.conflicts)
	} 
```

nit: simpler :)

testdata/src/go.uber.org/annotationparse/annotationparse.go
yuxincs: 👍

yuxincs (approved):
Love this unification!

Just a nit, but otherwise LGTM!

accumulation/analyzer.go
sonalmahajan15: Correct. Comment added.

inference/conflict.go
sonalmahajan15: Good point :)

PR #34, opened 2023-08-08
Refining error message text

sonalmahajan15:
This PR refines the error message text of produce and consume triggers in order to improve the overall UX.

[Depends on PR #31]

### DIFF
--- a/annotation/consume_trigger.go
+++ b/annotation/consume_trigger.go
@@ -132,7 +132,7 @@ func (p PtrLoad) Prestring() Prestring {
 type PtrLoadPrestring struct{}
 
 func (PtrLoadPrestring) String() string {
-	return "dereferenced"
+	return "`<EXPR>` dereferenced"
 }
 
 // MapAccess is when a map value flows to a point where it is indexed, and thus must be non-nil
@@ -155,7 +155,7 @@ func (i MapAccess) Prestring() Prestring {
 type MapAccessPrestring struct{}
 
 func (MapAccessPrestring) String() string {
-	return "keyed into"
+	return "map `<EXPR>` keyed into"
 }
 
 // MapWrittenTo is when a map value flows to a point where one of its indices is written to, and thus
@@ -177,7 +177,7 @@ func (m MapWrittenTo) Prestring() Prestring {
 type MapWrittenToPrestring struct{}
 
 func (MapWrittenToPrestring) String() string {
-	return "written to at an index"
+	return "map `<EXPR>` is written to at an index"
 }
 
 // SliceAccess is when a slice value flows to a point where it is sliced, and thus must be non-nil
@@ -198,12 +198,14 @@ func (s SliceAccess) Prestring() Prestring {
 type SliceAccessPrestring struct{}
 
 func (SliceAccessPrestring) String() string {
-	return "sliced into"
+	return "`<EXPR>` sliced into"
 }
 
 // FldAccess is when a value flows to a point where a field of it is accessed, and so it must be non-nil
 type FldAccess struct {
 	ConsumeTriggerTautology
+	Field  *types.Var
+	Method *types.Func
 }
 
 func (f FldAccess) String() string {
@@ -212,14 +214,30 @@ func (f FldAccess) String() string {
 
 // Prestring returns this FldAccess as a Prestring
 func (f FldAccess) Prestring() Prestring {
-	return FldAccessPrestring{}
+	fieldName, methodName := "", ""
+	if f.Field != nil {
+		fieldName = f.Field.Name()
+	} else if f.Method != nil {
+		methodName = f.Method.Name()
+	}
+
+	return FldAccessPrestring{
+		FieldName:  fieldName,
+		MethodName: methodName,
+	}
 }
 
 // FldAccessPrestring is a Prestring storing the needed information to compactly encode a FldAccess
-type FldAccessPrestring struct{}
+type FldAccessPrestring struct {
+	FieldName  string
+	MethodName string
+}
 
-func (FldAccessPrestring) String() string {
-	return "passed to a field access"
+func (f FldAccessPrestring) String() string {
+	if f.MethodName != "" {
+		return fmt.Sprintf("method `%s` of `<EXPR>` invoked", f.MethodName)
+	}
+	return fmt.Sprintf("field `%s` of `<EXPR>` accessed", f.FieldName)
 }
 
 // UseAsErrorResult is when a value flows to the error result of a function, where it is expected to be non-nil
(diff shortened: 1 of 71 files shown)


config/const.go
```diff
@@ -19,7 +19,7 @@ const StableRoundLimit = 5
 // BackpropTimeout is the timeout set for analysis of each function. This ensures build time SLAs.
 // NilAway should report an error if this timeout is ever hit, in order not to silently ignore any
 // functions due to this.
-const BackpropTimeout = 10 * time.Second
+const BackpropTimeout = 10000 * time.Second
```
lazaroclapp, line 22: Why are we increasing the timeout here to multiple hours? Leftover debugging change?

inference/conflict.go
```diff
@@ -97,7 +97,8 @@ func (n *nilFlow) String() string {
 	lastNonnilNodeIndex := len(n.nonnilPath) - 1
 	n.nonnilPath[lastNonnilNodeIndex].reason = fmt.Sprintf("%s (must be NONNIL)", n.nonnilPath[lastNonnilNodeIndex].reason)
 
-	return pathString(n.nilPath) + pathString(n.nonnilPath)
+	flow := pathString(n.nilPath) + pathString(n.nonnilPath)
+	return flow
```
lazaroclapp, line 101: Why this change?

util/util.go
```diff
@@ -452,7 +452,8 @@ func ExprToString(e ast.Expr, pass *analysis.Pass) string {
 	var buf bytes.Buffer
 	err := printer.Fprint(&buf, pass.Fset, e)
 	if err != nil {
-		panic(fmt.Sprintf("Failed to convert AST expression to string: %v\n", err))
+		// printer failed to convert AST expression to string, hence return empty string to avoid any panics downstream
+		return ""
```
lazaroclapp, line 456: When is this expected to happen?

config/const.go
sonalmahajan15: Yes 😄. Reverted.

inference/conflict.go
sonalmahajan15: It seems to be stemming from a rebasing issue.

annotation/consume_trigger.go
```diff
@@ -204,6 +204,8 @@ func (SliceAccessPrestring) String() string {
 // FldAccess is when a value flows to a point where a field of it is accessed, and so it must be non-nil
 type FldAccess struct {
 	ConsumeTriggerTautology
+	Field  *types.Var
+	Method *types.Func
```
yuxincs, line 208: IIUC, one and only one of these two must be nonnil, right? Can we make that explicit in the comments here?

Alternatively, we can just store `Sel types.Object` here and use type assertions to determine whether it is `*types.Var` or `*types.Func`

annotation/consume_trigger.go
```diff
@@ -212,14 +214,30 @@ func (f FldAccess) String() string {
 
 // Prestring returns this FldAccess as a Prestring
 func (f FldAccess) Prestring() Prestring {
-	return FldAccessPrestring{}
+	fieldName, methodName := "", ""
+	if f.Field != nil {
+		fieldName = f.Field.Name()
+	} else if f.Method != nil {
+		methodName = f.Method.Name()
+	}
```
yuxincs, line 222: panic otherwise?

inference/conflict.go
```diff
@@ -110,18 +110,18 @@ func pathString(nodes []node) string {
...
+	allNodes = append(allNodes, n.nilPath...)
+	allNodes = append(allNodes, n.nonnilPath...)
+
 	flow := ""
-	for _, nodes := range [...][]node{n.nilPath, n.nonnilPath} {
-		for _, nodeObj := range nodes {
-			flow += nodeObj.String()
-		}
+	for _, nodeObj := range allNodes {
+		flow += nodeObj.String()
```
yuxincs, line 119: Any reason for this change?

inference/conflict.go
```diff
@@ -69,16 +69,16 @@ func newNode(p annotation.Prestring, c annotation.Prestring) node {
...
-	} else if len(n.producerRepr) > 0 {
-		reasonStr = n.producerRepr
-	} else if len(n.consumerRepr) > 0 {
-		reasonStr = n.consumerRepr
+
+	if len(n.producerRepr) > 0 {
+		reasonStr += n.producerRepr + " "
+	}
+	if len(n.consumerRepr) > 0 {
+		reasonStr += n.consumerRepr
```
yuxincs, line 81: So this means if `n.producerRepr` is non-empty and `n.consumerRepr` is empty, we would result in a string `<producerRepr> ` (notice the trailing space).

Not that it matters much from UX perspective, but perhaps we do not want that anyways?

annotation/produce_trigger.go
```diff
@@ -333,28 +333,33 @@ func (UnassignedFld) Prestring() Prestring {
...
 
 func (UnassignedFldPrestring) String() string {
-	return "unassigned at struct initialization"
+	return "uninitialized"
 }
 
 // NoVarAssign is when a value is determined to flow from a variable that wasn't assigned to
 type NoVarAssign struct {
 	ProduceTriggerTautology
+	VarDecl *types.Var
```
yuxincs, line 342:
```suggestion
	VarObj *types.Var
```

I know this is done for consistency with the existing codebase. But I feel we should use `XXObj` instead of `XXDecl` for `*types.XX` since it refers to the object (entity) but not its declaration (which is an AST node). 

See more at https://github.com/golang/example/blob/master/gotypes/go-types.md

yuxincs (approved):
Great error message refinement!!

I have a few nits and a question, but none of them are blocking

annotation/consume_trigger.go
sonalmahajan15: Good idea. Went ahead with the latter option.

inference/conflict.go
sonalmahajan15: Short answer is it was causing errors being dropped.

Long answer: Single assertion conflict has only one producer and consumer. Now with the concatenation logic, this got combined in a single node which is stored in the `nonnilPath`. Thus keeping `nilPath` nil. In such cases , `for _, nodes := range [...][]node{n.nilPath, n.nonnilPath} {` was causing the combined slice to be nil, causing the loop to be skipped, dropping errors.

inference/conflict.go
sonalmahajan15: Good catch. Fixed.

annotation/consume_trigger.go
sonalmahajan15: Done

annotation/produce_trigger.go
sonalmahajan15: Right, changed here. But we should systematically change the entire codebase to follow the convention.

PR #36, opened 2023-08-09
[Follow-up of PR #24] Support function contract nonnil->nonnil for no infer mode

zzqatuber:
Follow up of PR #24 , support no infer mode for function contract nonnil->nonnil. This is the final PR of the stack.

For no infer mode, we annotate param and result at every call site of the contracted function. We duplicate param and return sites so they would not conflict with each other (which is done in #24 ). Theoretically we do not need to duplicate full triggers of contracted functions, but in order to keep the analysis phase the same between full and no infer mode, we still duplicate (done in #24) but we just ignore those duplicated triggers in the inference phase when we are in no infer mode.

Implementation:
- Parse inline annotations at call sites.
- Use the parsed annotations to check for conflicts.

### DIFF
--- a/annotation/full_trigger.go
+++ b/annotation/full_trigger.go
@@ -38,12 +38,21 @@ type FullTrigger struct {
 	// If this field is nil, it means the trigger is not a controlled trigger and the trigger will
 	// be activated all the time.
 	Controller *CallSiteParamAnnotationKey
+	// CreatedFromDuplication is true if the full trigger is created from duplicating another full
+	// trigger; otherwise false, which is also the default value for any normal full trigger.
+	CreatedFromDuplication bool
 }
 
 // Controlled returns true if this full trigger is controlled by a controller site; otherwise
 // returns false.
 func (t *FullTrigger) Controlled() bool {
 	return t.Controller != nil
+}
+
+// Duplicated returns true if the full trigger is created from duplicating another full trigger;
+// otherwise returns false.
+func (t *FullTrigger) Duplicated() bool {
+	return t.CreatedFromDuplication
 }
 
 // Pos returns the position for logging the error specified by the ConsumeTrigger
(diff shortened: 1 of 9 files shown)


annotation/full_trigger.go
```diff
@@ -46,6 +49,12 @@ func (t *FullTrigger) Controlled() bool {
 	return t.Controller != nil
 }
 
+// Duplicated returns true if the full trigger is created from duplicating another full trigger;
+// otherwise returns false.
+func (t *FullTrigger) Duplicated() bool {
+	return t.CreatedFromDuplication
+}
+
```
yuxincs, line 57: I think we can just use the exported field instead of having this method.

Encapsulation is good, but here I don't think it's really necessary :)

inference/engine.go
```diff
@@ -325,6 +325,11 @@ func (e *Engine) observeSiteExplanation(site primitiveSite, siteExplained Explai
 		}
 		e.conflicts.addConflict(newOverconstrainedConflict(site, trueExplanation, falseExplanation))
 
+		// Even though we have a conflict, we still need to make sure to activate any controlled
+		// triggers that are waiting on this site, so that we would not miss processing any
+		// triggers.
+		e.activateControlledTriggers(site, siteExplained)
+
```
yuxincs, line 332: Not saying we shouldn't do this, but just want to make sure this is a bugfix for the inference mode, right? (since this PR is on no infer mode).

yuxincs (approved):
This LGTM :)

I only have a nit and a question

inference/engine.go
zzqatuber: Yes you are right; it is a bugfix for inference mode! I don't want to open a PR for this single line, since this PR with its predecessors (#8 #24) is in a stack anyway.

annotation/full_trigger.go
zzqatuber: Good point, removed.

yuxincs (approved):
LGTM 💯 !

PR #38, opened 2023-08-10
Fix for getting the final consumer expression in the error message

sonalmahajan15:
This PR applies a fix for retrieving the correct final consumer expression from the point of conflict.

[Depends on PR #34]


sonalmahajan15:
This PR is now obsolete with the revised error message format. Hence, closing it without merging.

PR #39, opened 2023-08-11
Update CI configurations to use Go 1.20 and 1.21

yuxincs:
[Go 1.21 is released](https://go.dev/blog/go1.21), so this PR upgrades the versions we use in CI (from 1.19 and 1.20 to 1.20 and 1.21). Also, this PR makes it explicit in the README that we follow the same version support policy as the Go project: only last two major versions are supported.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: 1.20.x
+          go-version: 1.21.x
           cache: false
 
       - uses: golangci/golangci-lint-action@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [ "1.19.x", "1.20.x" ]
+        go: [ "1.20.x", "1.21.x" ]
```
sonalmahajan15, line 35:
```suggestion
        go: [ "1.20.x", "1.22.x" ]
```

sonalmahajan15 (approved):
LGTM :)

PR #40, opened 2023-08-18
Automatically infer function contract(nonnil->nonnil)

zzqatuber:
Automatically infer contract(nonnil->nonnil) for every function, by doing path-sensitive nilness predicate analysis for every variable and check if the contract holds at all return sites.


sonalmahajan15 (approved):
This PR had been reviewed internally with with all comments addressed. So approving it right away. LGTM :)

PR #42, opened 2023-08-18
Support contracts across packages.

zzqatuber:
Code changes of this PR includes the general changes that support cross-package for NilAway. However, that part is not migrated to GitHub yet so I can only rebase onto [zzq/function-contracts/general-nonnil](https://github.com/uber-go/nilaway/tree/zzq/function-contracts/general-nonnil). Ideally we should have another PR for the general changes to support cross-package between and rebase this PR to the PR.

What was done:
1. Pass around function contracts analyzed across packages.
2. Stop modifying the full triggers passed into inference engine; instead of having a new type Implication (which is exportable as package facts) to carry information from a FullTrigger and then adapt the engine to digest this new type.
3. Export and import contracts for full triggers of contracted functions across packages, in the form of implications.

Future work:
The current implementation does not work sometimes, because it creates a primitivizer in function analyzer but the position information this primitivizer obtains will be conflicting with those obtained from the primitivizer in accumulation analyzer.

Function analyzer cannot get the column number for every object from upstream, because unlike InferredMap exported in accumuulation analyzer, the Implication exported in function analyzer does not contain every exported object in upstream (We store only duplicable implications there)

Thus, a entirely new implementation is needed. We have to move all the logic of duplicating full triggers from function analyzer to accumalation analyzer (i.e., inference phase), so we maintain only one primitivizer and the position information (column number) of upstream objects can be obtained from upstream packages correctly.

PS:
Did not find integration tests here so I did not migrate integration cross-packge tests but they can be found in the internal version of this pull request.

### DIFF
--- a/accumulation/analyzer.go
+++ b/accumulation/analyzer.go
@@ -131,7 +131,7 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
 		// Incorporate assertions from this package one-by-one into the inferredAnnotationMap, possibly
 		// determining local and upstream sites in the process. This is guaranteed not to determine any
 		// sites unless we really have a reason they have to be determined.
-		engine.ObservePackage(assertionsResult.FullTriggers)
+		engine.ObservePackage(assertionsResult.FullTriggers, assertionsResult.Implications)
 		inferredMap, diagnostics = engine.InferredMapWithDiagnostics()
 
 	case inference.NoInfer:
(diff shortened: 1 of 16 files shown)


yuxincs:
The functionality of this PR should (at least partially) be implemented in PR #240, so I'm closing this.

If we need to further enhance cross-package support for contract, we'll use this PR as reference and properly attribute the contributions :)

Issue #44, opened 2023-08-23
Observe affiliations of embedded interfaces


sonalmahajan15:
Interfaces can be embedded as fields in a struct. This implies that the struct object can access its declared methods and that the embedded interface can refer to any object that implements it. Analyze the affiliation between the interface fields and structs properly.

 
```
type I interface {
   // nilable(x)
   foo(x *int) *int
}

type A struct{}

func (*A) foo(x *int) *int {
   return x
}

type B struct {
   I
}
func test1() *int {
   var i I = &B{}
   return i.foo(nil) // report error since B does not implement I.foo()
}

func test2() *int {
   var i I = &B{&A{}}
   return i.foo(nil) // report contravariance error at A.foo() definition
}
 ```

Issue #46, opened 2023-08-24
Add support for capturing affiliation in nested composite for slice


sonalmahajan15:
Affiliations analysis fails to capture nested composite literal for slice type. It does not report any error for the below test case.

```
type A1 struct{}

func (A1) foo(x *int) *int {
	return x
}

func test() {
	var v [][]I = [][]I{{&A1{}}}
	_ = *(v[0][0].foo(nil)) // nil panic here
}
```

ccuetoh:
## Good First Issue - Help welcomed!

**Why this is a good first issue:**
The affiliation analyzer already records concrete types that flow into interface typed variables, and this gap is limited to types stored through composite literals nested inside slices. The work is to walk nested slice and array composite literals when they are assigned and record their elements the same way simple assignments already are. Once recorded, the existing interface method handling does the rest, and the affiliation tests show the expected structure for new cases.

PR #47, opened 2023-08-24
Support for handling interface fields in affiliations analysis

### LINKED ISSUE #43: Observe affiliations of embedded structs
A common Go pattern is to "embed" one struct into another - automatically making the latter implement every interface the former implemented. The affiliations analyzer does not take this into account, leading to false negatives. For example, the following test fails to produce the diagnostic it should:

```
type MyI interface {
   //nilable(x)
   foo(x *int) int
}

type S struct{}

//nonnil(x)
func (S) foo(x *int) int { //want "nilable value could be passed as param"
   return *x
}

type T struct {
   S
}

func bar() int {
   var i MyI
   i = T{}
   return i.foo(nil)
} 
```
If the line `i = T{}` were replaced with `i = S{}` then the appropriate error would be thrown.

sonalmahajan15:
This PR adds support for tracking an interface field of a struct as an affiliation site. The interface field could be embedded or declared as a normal field.

[closes #43 #44 ]

### DIFF
--- a/assertion/affiliation/affiliation.go
+++ b/assertion/affiliation/affiliation.go
@@ -181,14 +181,14 @@ func (a *Affiliation) computeTriggersForCastingSites(pass *analysis.Pass, upstre
 					switch nodeType := node.Type.(type) {
 					case *ast.ArrayType:
 						// A slice (or array) declared of type interface, and initialized with a struct
-						// e.g., var i I = []I{&S{}}
+						// e.g., _ = []I{&S{}}
 						lhsType := util.TypeOf(pass, nodeType.Elt)
 						for _, elt := range node.Elts {
 							appendTypeToTypeTriggers(lhsType, util.TypeOf(pass, elt))
 						}
 					case *ast.MapType:
 						// Key, value, or both of a map declared of type interface, and initialized with a struct
-						// e.g., var i I = map[int]I{0: &S{}}
+						// e.g., _ = map[int]I{0: &S{}}
 						keyType := util.TypeOf(pass, nodeType.Key)
 						valueType := util.TypeOf(pass, nodeType.Value)
 						for _, elt := range node.Elts {
@@ -205,11 +205,11 @@ func (a *Affiliation) computeTriggersForCastingSites(pass *analysis.Pass, upstre
 						for i, elt := range node.Elts {
 							var lhsType, rhsType types.Type
 							if kv, ok := elt.(*ast.KeyValueExpr); ok {
-								// In this case the initialization is key-value based. E.g. a = &A{f: &B{}}
+								// In this case the initialization is key-value based. E.g. s = &S{t: &T{}}
 								lhsType = util.TypeOf(pass, kv.Key)
 								rhsType = util.TypeOf(pass, kv.Value)
 							} else {
-								// In this case the initialization is serial. E.g. a = &A{&B{}}
+								// In this case the initialization is serial. E.g. s = &S{&T{}}
 								if sObj := util.TypeAsDeeplyStruct(util.TypeOf(pass, node)); sObj != nil {
 									lhsType = sObj.Field(i).Type()
 									rhsType = util.TypeOf(pass, elt)
--- a/assertion/affiliation/affiliation.go
+++ b/assertion/affiliation/affiliation.go
@@ -63,7 +63,7 @@ func (a *Affiliation) extractAffiliations(pass *analysis.Pass) {
 
 	// populate upstreamCache by importing entries passed from upstream packages
 	facts := pass.AllPackageFacts()
-	if facts != nil && len(facts) > 0 {
+	if len(facts) > 0 {
 		for _, f := range facts {
 			switch c := f.Fact.(type) {
 			case *AffliliationCache:
--- a/assertion/affiliation/affiliation.go
+++ b/assertion/affiliation/affiliation.go
@@ -63,7 +63,7 @@ func (a *Affiliation) extractAffiliations(pass *analysis.Pass) {
 
 	// populate upstreamCache by importing entries passed from upstream packages
 	facts := pass.AllPackageFacts()
-	if len(facts) > 0 {
+	if facts != nil && len(facts) > 0 {
 		for _, f := range facts {
 			switch c := f.Fact.(type) {
 			case *AffliliationCache:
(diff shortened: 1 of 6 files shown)


assertion/function/assertiontree/backprop_util.go
```diff
@@ -719,7 +719,7 @@ func exprAsDeepProducer(rootNode *RootAssertionNode, expr ast.Expr) annotation.P
 	if len(parsedExpr) > 1 {
 		panic("multiply returning function passed where a deep producer is expected - tuple types are not deep")
 	}
-	if len(parsedExpr) == 0 || !parsedExpr[0].IsDeep() {
+	if len(parsedExpr) == 0 || !parsedExpr[0].IsDeep() || parsedExpr[0].GetDeep() == nil {
```
sonalmahajan15, line 722: (fix for NilAway reported error that was reported due the support added in this PR)

assertion/function/assertiontree/assertion_node.go
```diff
@@ -93,7 +93,7 @@ func (n *assertionNodeCommon) SetConsumeTriggers(triggers []*annotation.ConsumeT
 
 // nilable(node, result 0)
 func (n *assertionNodeCommon) Root() *RootAssertionNode {
-	if n == nil {
+	if n == nil || n.parent == nil {
```
sonalmahajan15, line 96: (fix for NilAway reported error that was reported due the support added in this PR)

assertion/function/assertiontree/util.go
```diff
@@ -61,6 +61,10 @@ func RootNodeSlicesString(name string, rootNodes []*RootAssertionNode) string {
 // this combines each of the special-cased deep nilability introspectors to give a method that
 // determines a deep nilability trigger for an arbitrary assertion node
 func deepNilabilityTriggerOf(node AssertionNode) annotation.ProducingAnnotationTrigger {
+	if node == nil {
```
sonalmahajan15, line 64: (fix for NilAway reported error that was reported due the support added in this PR)

assertion/function/assertiontree/backprop_util.go
yuxincs: This is a wontfix false positive since `.IsDeep() => .GetDeep() != nil`, right? 

(honestly, I feel we could probably just refactor the struct to get rid of `IsDeep()`, but that should definitely happen in separate PRs

assertion/affiliation/affiliation.go
```diff
@@ -178,48 +178,51 @@ func (a *Affiliation) computeTriggersForCastingSites(pass *analysis.Pass, upstre
...
-					// for keys and values. For instance, if the type of value of the map is an interface
-					// but a struct is added to the map.
-					if mpType, ok := nodeType.(*types.Map); ok {
-						elemType := mpType.Elem()
-						keyType := mpType.Key()
-						// Iterate through each element in the composite
+					switch nodeType := node.Type.(type) {
+					case *ast.ArrayType:
+						// A slice (or array) declared of type interface, and initialized with a struct
+						// e.g., _ = []I{&S{}}
```
yuxincs, line 184: Can we add some comments stating the current limitations (i.e., https://github.com/uber-go/nilaway/issues/46)? :) I had a feeling that we may need help from the type-checker in the future since Go allows you omit some type parts on syntactical level....

assertion/function/assertiontree/util.go
yuxincs: I wonder where `deepNilabilityTriggerOf` ever gets passed a NILABLE node 🤔

yuxincs (approved):
Most of the logic is already reviewed and approved internally. 

I only have a couple of nits and questions, but none of them are blocking. 👍

assertion/affiliation/affiliation.go
sonalmahajan15: Good point. Added a comment linking the issue.

assertion/affiliation/affiliation.go
```diff
@@ -178,48 +178,51 @@ func (a *Affiliation) computeTriggersForCastingSites(pass *analysis.Pass, upstre
...
+							}
+							if lhsType != nil && rhsType != nil {
+								appendTypeToTypeTriggers(lhsType, rhsType)
 							}
 						}
 					}
 				case *ast.FuncLit:
 					// TODO: Nilability analysis support for anonymous functions is currently not
-					//       implemented (tracked in , so here we completely skip
+					//       implemented (tracked in PROGSYS-559), so here we completely skip
```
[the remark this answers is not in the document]
sonalmahajan15: Ah, good catch. Updated to reflect the newly created issue #52.

assertion/function/assertiontree/assertion_node.go
sonalmahajan15: > Wait....isn't the original logic incorrect? we would recursively go to n.parent until n == nil, and then return a nil instead. So Root() always return nil in all cases.

You are right that this function would result in recursively going up to the root of the assertion tree, which is of type `RootAssertionNode`. However, `RootAssertionNode` overrides the `Root()` method to return itself, which is why `nil` was never being returned from here.

> A quick search reveals that this function isn't called anywhere, can we simply remove this function then?

I actually see it being used from a lot of places, such as
[https://sourcegraph.uberinternal.com/code.uber.internal/go-code/-/blob/src/code.uber.internal/infra/progsys/nilaway/assertion/function/assertiontree/var_assertion_node.go#L53](https://sourcegraph.uberinternal.com/code.uber.internal/go-code/-/blob/src/code.uber.internal/infra/progsys/nilaway/assertion/function/assertiontree/var_assertion_node.go#L53)
[https://sourcegraph.uberinternal.com/code.uber.internal/go-code/-/blob/src/code.uber.internal/infra/progsys/nilaway/assertion/function/assertiontree/var_assertion_node.go#L71](https://sourcegraph.uberinternal.com/code.uber.internal/go-code/-/blob/src/code.uber.internal/infra/progsys/nilaway/assertion/function/assertiontree/var_assertion_node.go#L71)

assertion/function/assertiontree/backprop_util.go
sonalmahajan15: Correct, a "wontfix" false positive. I remember I had a implemented a refactor a while ago internally on lines similar to what you are suggestions for solving this problem, I'll dig it up and create a new PR.

assertion/function/assertiontree/util.go
sonalmahajan15: Never directly. It is a false positive from a context sensitivity point of view: [assertiontree/util.go:L106](https://sourcegraph.uberinternal.com/code.uber.internal/go-code/-/blob/src/code.uber.internal/infra/progsys/nilaway/assertion/function/assertiontree/util.go#L106) sets parent to be nil, and [assertiontree/index_assertion_node.go#L31](https://sourcegraph.uberinternal.com/code.uber.internal/go-code/-/blob/src/code.uber.internal/infra/progsys/nilaway/assertion/function/assertiontree/index_assertion_node.go#L31), for instance, calls `deepNilabilityTriggerOf` with `i.Parent()`, which could be nil from the previous finding.

assertion/function/assertiontree/assertion_node.go
```diff
@@ -93,7 +93,7 @@ func (n *assertionNodeCommon) SetConsumeTriggers(triggers []*annotation.ConsumeT
 
 // nilable(node, result 0)
```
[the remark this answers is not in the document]
sonalmahajan15: Done :)

PR #49, opened 2023-08-28
Support for pkg.Const as a stable expression

### LINKED ISSUE #48: False positive reported for use of pkg.Const in index expression
The following code produces a false positive. `pkg.Const` is a string constant defined in the package `pkg`. Although, it is checked that the value of `mp[pkg.Const]` is not nil before dereferencing, NilAway still reports an error.
```
func foo(mp map[string]*string) {
      if mp == nil || mp[pkg.Const] == nil || *mp[pkg.Const] == "" {  // error reported for the dereference of mp[key]
		// no op
      }
}
```

sonalmahajan15:
NilAway already regards locally defined constants as stable expressions. However, it does not extend that behavior to constants defined in other packages accessed via selector expression `pkg.Const`. This PR adds support for the same by augmenting the logic of stable expressions.

[Closes #48]

### DIFF
--- a/assertion/function/assertiontree/root_assertion_node.go
+++ b/assertion/function/assertiontree/root_assertion_node.go
@@ -1175,8 +1175,16 @@ func (r *RootAssertionNode) isStable(expr ast.Expr) bool {
 			return true
 		}
 
-		// TODO: check for function names and user-declared constants
+		// package is considered a special case of ident to suppport selector expressions used to access stable
+		// expressions, such as constants declared in another package (e.g., pkg.Const)
+		if r.isPkgName(expr) {
+			return true
+		}
+
+		// TODO: check for function names
 		return false
+	case *ast.SelectorExpr:
+		return r.isStable(expr.Sel) && r.isStable(expr.X)
 	default:
 		return false
 	}
@@ -1221,7 +1229,8 @@ func (r *RootAssertionNode) eqStable(left, right ast.Expr) bool {
 			// if the two identifiers are special values, just check them for string equality
 			if (r.isNil(left) && r.isNil(right)) ||
 				(r.isBuiltIn(left) && r.isBuiltIn(right)) ||
-				(r.isConst(left) && (r.isConst(right))) {
+				(r.isConst(left) && (r.isConst(right))) ||
+				(r.isPkgName(left) && r.isPkgName(right)) {
 				return left.Name == right.Name
 			}
 			rightVarObj, rightOk := r.ObjectOf(right).(*types.Var)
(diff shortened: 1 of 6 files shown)


testdata/src/go.uber.org/maps/maps.go
```diff
@@ -692,3 +698,56 @@ func testExplicitBool(mp map[int]*int, i int) *int {
 	}
 	return &i
 }
+
+// tests for checking index expressions as constants both built-in and user defined (declared locally and in another package)
+// nonnil(mp, mp[])
+func testIndexAsUserDefinedConst(mp map[string]*string, i int) string {
```
yuxincs, line 704: 👍 (why don't we create a new package, maybe `consts`, instead of having it reside in `maps`?)

testdata/src/go.uber.org/maps/lib/const.go
```diff
@@ -0,0 +1,3 @@
+package lib
+
+const MyStrConst = "xyz"
```
yuxincs, line 3: Can we have a global variable here and corresponding test cases as well (such that it should not be considered stable)? Or is it already covered somewhere else (even so, I feel we can just add a simple case for sake of completeness)?

assertion/function/assertiontree/root_assertion_node.go
```diff
@@ -1175,8 +1175,16 @@ func (r *RootAssertionNode) isStable(expr ast.Expr) bool {
 			return true
 		}
 
-		// TODO: check for function names and user-declared constants
+		// package is considered a special case of ident to suppport selector expressions used to access stable
+		// expressions, such as constants declared in another package (e.g., pkg.Const)
+		if r.isPkgName(expr) {
+			return true
+		}
```
yuxincs, line 1182: Can we just combine the cases into a single `if` while we are at it? 😃

yuxincs (approved):
This was discovered some time ago, but we never got a chance to implement the support. Thanks a lot 🚀 !!

I only have a couple of nits and an additional test case request 👍

assertion/function/assertiontree/root_assertion_node.go
sonalmahajan15: Sure, done :)

testdata/src/go.uber.org/maps/maps.go
sonalmahajan15: Fair point. Done.

testdata/src/go.uber.org/maps/lib/const.go
sonalmahajan15: That's fair. I have added a test case here.

PR #50, opened 2023-08-30
Fix multi-package inference algorithm which incorrectly uses token.Pos for injectivity

yuxincs:
Our multi-package inference algorithm is broken in practice for incremental build systems like bazel:

In order to incur minimal overhead to storage, we store a stripped down version of annotation site (i.e., `inference.primitiveSite`) in the inference map that compactly describes the annotation site. In this struct, we are also storing `token.Pos` as position information of the site for injectivity.

(`token.Pos` is simply an integer that cleverly and compactly encodes the position information in a given file set, so that you don’t have to store `(filename, line, column)` inside each AST node to reduce memory consumption. Whenever you need to access this information, you can convert by querying the file set `Fset.Position(p Pos)` . To put differently, `token.Pos` means nothing without the `Fset` context.)

Now, let's say we have analyzed upstream package, exported `NilableValue` in the inferred map (which includes a `token.Pos` that only makes sense in the context of the upstream package’s `Fset` ). When analyzing the downstream package, we will first import the `NilableValue` annotation site (with a `token.Pos` that is only meaningful during analysis of the upstream package) and its value in our map, and then check the dereference `upstream.NilableValue`. However, the `token.Pos` for the dereference point is for the downstream package and could be completely different - we won’t find it in the map, no conflicts => no errors.

In this PR, we replace `token.Pos` with `token.Position` (i.e., a tuple of filename, offset, line, column) to make it package-independent. However, this still does not fix the whole problem since we won't have complete position information about upstream sites when analyzing downstream packages. More specifically, the upstream file in the `analysis.Pass.Fset` will not have complete line and column  information. Instead, the [importer](https://cs.opensource.google/go/x/tools/+/refs/tags/v0.7.0:go/internal/gcimporter/bimport.go;l=375-385;drc=c1dd25e80b559a5b0e8e2dd7d5bd1e946aa996a0;bpv=0;bpt=0) injects 65535 fake "lines" into the file, and the object we get for dereference `upstream.NilableValue` will still contain completely different position information due to this hack (in practice, we have observed that the lines for the objects are correct, but not others). This leads to mismatches in our inference engine: we still cannot find nilabilities of upstream objects in the imported `InferredMap` since their Position fields in the primitive sites are different. 

Therefore, in addition to the package-independent `token.Position`, we have also added an `ObjectPath` field to the exported site, that is an opaque name that identifies a `types.Object` relative to its package (see [objectpath](https://pkg.go.dev/golang.org/x/tools/go/types/objectpath) package for more details). This is then used to match the corresponding upstream objects in downstream analysis, and fix the positions information according to the imported positions from upstream analyses.

Implementation-wise, this PR contains the following changes:

* We replace `token.Pos` with `token.Position` and add an `ObjectPath` field in primitive sites.
* We add a `primitivizer` struct that handles caching, and fixing logic for upstream objects. Whenever we need to convert full trigger or annotation sites to their primitive forms, we _must_ call methods on this struct to get the correct primitive sites.
* We trim the bazel-specific prefix of `filename` in `token.Position` in primitive sites since bazel will create a sandbox with a random prefix for each build.

Depends on PR #63

### DIFF
--- a/diagnostic/engine.go
+++ b/diagnostic/engine.go
@@ -18,9 +18,9 @@
 package diagnostic
 
 import (
+	"fmt"
 	"go/token"
 	"os"
-	"fmt"
 	"path/filepath"
 
 	"go.uber.org/nilaway/annotation"
--- a/diagnostic/engine.go
+++ b/diagnostic/engine.go
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package diagnostic hosts the diagnostic engine, which is responsible for collecting the
+// conflicts from annotation-based checks (no-infer mode) and/or inference (full-infer mode) and
+// generating user-friendly diagnostics from those conflicts.
 package diagnostic
 
 import (
--- a/diagnostic/engine.go
+++ b/diagnostic/engine.go
@@ -46,7 +46,7 @@ func (e *Engine) Diagnostics(grouping bool) []analysis.Diagnostic {
 	}
 
 	// build diagnostics from conflicts
-	var diagnostics []analysis.Diagnostic
+	diagnostics := make([]analysis.Diagnostic, 0, len(conflicts))
 	for _, c := range conflicts {
 		diagnostics = append(diagnostics, analysis.Diagnostic{
 			Pos:     c.pos,
--- a/diagnostic/engine.go
+++ b/diagnostic/engine.go
@@ -18,21 +18,79 @@
 package diagnostic
 
 import (
+	"go/token"
+	"os"
+	"fmt"
+	"path/filepath"
+
 	"go.uber.org/nilaway/annotation"
 	"go.uber.org/nilaway/inference"
 	"go.uber.org/nilaway/util"
 	"golang.org/x/tools/go/analysis"
 )
 
+// fileInfo bundles the token.File object and auxiliary information about it, e.g., whether it is
+// a fake file (i.e., imported from archive), for uses in primitivizer.
+type fileInfo struct {
+	file   *token.File
+	isFake bool
+}
+
 // Engine is the main engine for generating diagnostics from conflicts.
 type Engine struct {
 	pass      *analysis.Pass
 	conflicts []conflict
+	// files maps the file name (modulo the possible build-system prefix) to the token.File object
+	// for faster lookup when converting correct upstream position back to local token.Pos for
+	// reporting purposes.
+	files map[string]fileInfo
 }
 
 // NewEngine creates a new diagnostic engine.
 func NewEngine(pass *analysis.Pass) *Engine {
-	return &Engine{pass: pass}
+	// Find the current working directory (e.g., random sandbox prefix) for trimming the file names.
+	cwd, err := os.Getwd()
+	if err != nil {
+		panic(fmt.Sprintf("cannot get current working directory: %v", err))
+	}
+
+	// Iterate all files within the Fset (which includes upstream and current-package files), and
+	// store the mapping between its file name (modulo the possible build-system prefix) and the
+	// token.File object. This is needed for converting correct upstream position back to local
+	// incorrect token.Pos for error reporting purposes.
+	files := make(map[string]fileInfo)
+	pass.Fset.Iterate(func(file *token.File) bool {
+		name, err := filepath.Rel(cwd, file.Name())
+		if err != nil {
+			// For files that are not in the execroot (e.g., stdlib files start with "$GOROOT", and
+			// upstream files that do not have the build-system prefix), we can simply use the
+			// original file name.
+			name = file.Name()
+		}
+
+		// The file will be fake (conceptually "\n" * 65535) if it is imported from archive. So we
+		// check if there are any gaps between the line starts to determine if the file is fake.
+		// TODO: Go 1.21 introduced a new "token.File.Lines()" API to directly get the underlying
+		//  lines slice, which should allow faster checks since calling "token.File.LineStart"
+		//  repeatedly is slow due to locks/unlocks.
+		isFake := true
+		var prev token.Pos
+		for i := 1; i <= file.LineCount(); i++ {
+			p := file.LineStart(i)
+			if prev.IsValid() && p-prev > 1 {
+				isFake = false
+				break
+			}
+			prev = p
+		}
+		files[name] = fileInfo{
+			file:   file,
+			isFake: isFake,
+		}
+		return true
+	})
+
+	return &Engine{pass: pass, files: files}
 }
 
 // Diagnostics generates diagnostics from the internally-stored conflicts. The grouping parameter
@@ -70,7 +128,7 @@ func (e *Engine) AddSingleAssertionConflict(trigger annotation.FullTrigger) {
 
 // AddOverconstraintConflict adds a new overconstraint conflict to the engine.
 func (e *Engine) AddOverconstraintConflict(nilReason, nonnilReason inference.ExplainedBool) {
-	c := conflict{}
+	flow := nilFlow{}
 
 	// Build nil path by traversing the inference graph from `nilReason` part of the overconstraint failure.
 	// (Note that this traversal gives us a backward path from point of conflict to the source of nilability. Hence, we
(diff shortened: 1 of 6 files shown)


inference/primitive.go
```diff
@@ -38,67 +42,300 @@ type primitiveFullTrigger struct {
...
+				panic(fmt.Sprintf(
+					"conflicting position information on upstream object %q: existing: %v, got: %v",
+					objRepr, existing, site.Position))
+			}
+			upstreamObjPositions[objRepr] = site.Position
+			return true
+		})
+	}
+
+	// Find the current working directory (e.g., random sandbox prefix) for trimming the file names.
```
lazaroclapp, line 178: Interesting, does bazel set the CWD to something inside `bazel-out/` regardless of where the `bazel build` command is run from? How about other build systems? Feels like, in general, making all paths relative to CWD for most build systems would cause issues when you run the build command from different directories...

Maybe we want to have this logic iff cwd includes `bazel-out/`? Or does that directory name change?

inference/primitive.go
```diff
@@ -38,67 +42,300 @@ type primitiveFullTrigger struct {
...
+	// to primitiveSite, which can be used to uniquely identify an exported object relative to the
+	// package. Then, we simply cache the correct position information when importing
+	// InferredMaps, since the positions collected in the upstream analysis are always correct.
+	// Later when querying upstream objects in downstream analysis, we can look up the cache and
+	// fill in the correct position in the returned primitive site instead.
+
+	// Create a cache for upstream object positions.
+	upstreamObjPositions := make(map[string]token.Position)
+	for _, pkgFact := range pass.AllPackageFacts() {
+		importedMap, ok := pkgFact.Fact.(*InferredMap)
```
lazaroclapp, line 157: We are now scanning for all `InferredMap` facts exported by NilAway in upstream dependencies multiple times, right? Is this fast enough or should we be doing this wherever else we load these maps?

lazaroclapp (commented):
Ok, some minor nits/questions below, but overall I think I get and it LGTM.

I assume all test cases for this have to be internal integration tests since this only reproduces with bazel+nogo and we aren't building on bazel on OSS for now?

inference/engine.go
```diff
@@ -240,15 +245,15 @@ func (e *Engine) buildFromSingleFullTrigger(trigger annotation.FullTrigger) {
 	case pKind == annotation.Always && cKind == annotation.Always:
 		// Producer always produces nilable value -> consumer always consumes nonnil value.
 		// We simply generate a failure for this case.
-		e.conflicts.addConflict(newSingleAssertionConflict(e.pass, trigger))
+		e.conflicts.addConflict(newSingleAssertionConflict(trigger.Consumer.Expr.Pos(), e.primitive.fullTrigger(trigger), trigger))
```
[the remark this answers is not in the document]
yuxincs: > Is Consumer guaranteed to be on the current target here?

Yes, all full triggers are generated for the current package, and we are using `Consumer.Expr.Pos()`, which is the pos of the AST node, double ensuring the pos here is for the local package only.

But due to our current architecture changes (new error format + separate diagnostic engine), this comment no longer applies :)

inference/primitive.go
```diff
@@ -33,72 +37,212 @@ import (
...
+
+	return &primitivizer{
+		pass:                 pass,
+		upstreamObjPositions: upstreamObjPositions,
+		curDir:               cwd,
+	}
+}
+
+// fullTrigger returns the primitive version of the full trigger.
+func (p *primitivizer) fullTrigger(trigger annotation.FullTrigger) primitiveFullTrigger {
```
sonalmahajan15, line 178: Could we rename `fullTrigger()` to something more intuitive (e.g., `simplifiedFullTrigger()`)? I think it could be a little confusing to call this `fullTrigger`, when we are actually returning its simplified or primitive version.

inference/primitive.go
sonalmahajan15: Does the comment still hold? If yes, did the performance evaluation succeed internally?

sonalmahajan15 (approved):
Overall, looks good to me. A few nits/questions/suggestions listed below, but none of them are blockers.

inference/primitive.go
yuxincs: Yes, our internal performance evaluation shows the build time overhead increases from 2% to ~2.9%. However, this increase is mostly due to more sites being correctly connected and more triggers being generated, instead of the extra map access here 😃

inference/engine.go
```diff
@@ -277,10 +280,10 @@ func (e *Engine) buildFromSingleFullTrigger(trigger annotation.FullTrigger) {
 		if pSite == nil || cSite == nil {
 			panic("trigger is conditional but the underlying site is nil")
 		}
-		producer := newPrimitiveSite(pSite, pKind == annotation.DeepConditional)
-		consumer := newPrimitiveSite(cSite, cKind == annotation.DeepConditional)
+		producer := e.primitive.site(pSite, pKind == annotation.DeepConditional)
+		consumer := e.primitive.site(cSite, cKind == annotation.DeepConditional)
 
-		e.observeImplication(producer, consumer, primitiveAssertion)
+		e.observeImplication(producer, consumer, e.primitive.fullTrigger(trigger))
```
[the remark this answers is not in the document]
sonalmahajan15: Hmm, I see. We can leave it as it is. It was anyway nit, not a big deal :)

inference/primitive.go
sonalmahajan15: I see your point. I would still prefer to rename the method to make it more intuitive and decouple from the `full trigger` connotation  that exists otherwise in the code. But this is only a small nit, and can be ignored :)

inference/primitive.go
lazaroclapp: Makes sense. I suspect one day we will run into yet another analysis driver that breaks this in some other way (buck.v2?), but it's a bridge we can cross when we get to it.

inference/primitive.go
```diff
@@ -33,72 +37,212 @@ import (
...
-	if s.IsDeep {
-		deepStr = "Deep "
+// toPosition returns the correct position information for the given pos, removing sandbox prefix
+// if any.
+func (p *primitivizer) toPosition(pos token.Pos) token.Position {
+	// Generated files contain "//line" directives that point back to the original source file
+	// for better error reporting, and PositionFor supports reading that information and adjust
+	// the position accordingly. However, those source files are never truly analyzed, meaning
+	// the downstream analysis will try to look for the generated files instead of the source
+	// files. Therefore, here we use the unadjusted position instead.
```
lazaroclapp, line 241: I get an idea of what this comment is about, but not quite the consequences of passing `adjusted=true` here instead... what would that cause? (Just out of curiosity. I assume the choice is correct, but the explanation isn't super easy to follow for me now that I forgot so much of the context of our discussions about this PR :) )

lazaroclapp (approved):
Did a pass. With the exception of one question on how file sets are shared between analyzers, all my comments are basically comments-on-comments (and, after thinking for a bit, I am undecided on the naming convention issue Sonal raised, so you two figure that one out 🙂). Overall, this LGTM!

diagnostic/engine.go
```diff
@@ -18,21 +18,79 @@
...
+	// files maps the file name (modulo the possible build-system prefix) to the token.File object
+	// for faster lookup when converting correct upstream position back to local token.Pos for
+	// reporting purposes.
+	files map[string]fileInfo
 }
 
 // NewEngine creates a new diagnostic engine.
 func NewEngine(pass *analysis.Pass) *Engine {
-	return &Engine{pass: pass}
+	// Find the current working directory (e.g., random sandbox prefix) for trimming the file names.
```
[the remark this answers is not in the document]
yuxincs: Yes! Updated!

inference/primitive.go
yuxincs: I see. Somehow I still prefer the current name given that `primitive.simplifiedFullTrigger` reads a little confusing, `primitive.primitiveFullTrigger` is stuttering, and `primitive.ToPrimitiveFullTrigger` is a bit unnecessarily long.

But I get your point! Since this `primitivizer` struct and logic is only within `inference` package (unexported), and there is no immediate plan to export it. I feel it's relatively safe to keep it this way. If we ever need to export it in the future, we can make it clearer by adding some prefixes 😃 

Hope that also makes sense to you

PR #54, opened 2023-09-08
Enable struct object creation to be safe without struct init analysis

### LINKED ISSUE #53: Fix false positive in struct creation
The following code reports a false positive in the absence of struct initialization enabled.
```
type A struct {
     f int
}

func foo() {
    var x A
    y := &x
    print(y.f) // error reported here since `x` is incorrectly considered uninitialized
}
```

sonalmahajan15:
This PR allows struct object creation `var s S` to be considered safe (i.e., nonnil), avoiding reporting of false positives due to incorrect marking of `s` as uninitialized. This support was already added in NilAway, but was guarded by the struct init flag. This PR removes that restriction, allowing this support to be generally available.

[Closes #53]

### DIFF
--- a/assertion/function/assertiontree/var_assertion_node.go
+++ b/assertion/function/assertiontree/var_assertion_node.go
@@ -66,8 +66,8 @@ func (v *varAssertionNode) DefaultTrigger() annotation.ProducingAnnotationTrigge
 	//  preprocessing phase
 	if !util.TypeIsDeeplyPtr(v.decl.Type()) {
 		if structType := util.TypeAsDeeplyStruct(v.decl.Type()); structType != nil {
-			builtExpr := v.BuildExpr(v.Root().Pass(), nil)
 			if v.Root().functionContext.isDepthOneFieldCheck() {
+				builtExpr := v.BuildExpr(v.Root().Pass(), nil)
 				v.Root().addProductionForVarFieldNode(v, builtExpr)
 			}
 			return annotation.ProduceTriggerNever{} // indicating that the struct object itself is not nil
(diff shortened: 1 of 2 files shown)


assertion/function/assertiontree/var_assertion_node.go
```diff
@@ -63,14 +63,14 @@ func (v *varAssertionNode) DefaultTrigger() annotation.ProducingAnnotationTrigge
...
 	// TODO: below logic won't be required once we standardize the expression `var s S` by replacing it with `S{}` in the
-	//  preprocessing phase after  is implemented
-	if v.Root().functionContext.isDepthOneFieldCheck() {
-		if !util.TypeIsDeeplyPtr(v.decl.Type()) {
-			if structType := util.TypeAsDeeplyStruct(v.decl.Type()); structType != nil {
-				builtExpr := v.BuildExpr(v.Root().Pass(), nil)
+	//  preprocessing phase
+	if !util.TypeIsDeeplyPtr(v.decl.Type()) {
+		if structType := util.TypeAsDeeplyStruct(v.decl.Type()); structType != nil {
+			builtExpr := v.BuildExpr(v.Root().Pass(), nil)
```
yuxincs, line 69: Should we move this inside the if check below (or just remove this temp var since the function call is clear enough)?

yuxincs (approved):
LGTM! I just have a minor nit, appreciate the comprehensive test cases!

assertion/function/assertiontree/var_assertion_node.go
sonalmahajan15: Ah, good point. Done.

PR #60, opened 2023-09-26
Fix false positive due to local variable deep assign

### LINKED ISSUE #59: Deep assign to local variables causes false positives
NilAway currently does not allow deep nilability assignment to local variables, which results in false positives.

```
func testDeepAssignNil() {
   m := make(map[int]*string)
   m[0] = nil // FP: error reported here not allowing nil to be assigned
   if v, ok := m[0]; ok && v != nil {
       _ = *v
   }
} 
```

sonalmahajan15:
NilAway puts a restriction on deep nilable assignment to local variables. This can result in false positives, if the dereferences are properly guarded. This PR lifts this restriction and allows deep nilability into local variables, thereby addressing the false positives.

[closes #59]
[depends on #82 ]

### DIFF
--- a/annotation/key.go
+++ b/annotation/key.go
@@ -455,7 +455,7 @@ func (gk *GlobalVarAnnotationKey) String() string {
 	return fmt.Sprintf("Global Variable %s", gk.VarDecl.Name())
 }
 
-// LocalVarAnnotationKey allows the Lookup of a global variable's annotations in the Annotation Map
+// LocalVarAnnotationKey allows the Lookup of a local variable's annotations in the Annotation Map
 type LocalVarAnnotationKey struct {
 	VarDecl *types.Var
 }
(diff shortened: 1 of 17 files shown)


annotation/key.go
```diff
@@ -455,6 +455,38 @@ func (gk *GlobalVarAnnotationKey) String() string {
 	return fmt.Sprintf("Global Variable %s", gk.VarDecl.Name())
 }
 
+// LocalVarAnnotationKey allows the Lookup of a global variable's annotations in the Annotation Map
```
yuxincs, line 458:
```suggestion
// LocalVarAnnotationKey allows the Lookup of a local variable's annotations in the Annotation Map
```

right?

yuxincs (approved):
Nice find with a simple but great fix!

I only have a suggestion on a doc fix 😄

annotation/key.go
sonalmahajan15: Yes! Good catch :)

PR #62, opened 2023-09-28
Separate the diagnostic processing engine from inference package

### LINKED ISSUE #55: Refactor error reporting logic to a new package
Discussion link: https://github.com/uber-go/nilaway/pull/28/files#r1295204644

yuxincs:
We now have enough logic in our conflict handling logic (i.e., collecting conflicts and generating user-friendly diagnostics from them). For easier maintenance and future development, this PR separates the logic out from the inference package.

After this PR, we will have three major components in NilAway: analyzer engine(that generates full triggers), inference engine (that generates inferred maps and conflicts), and the diagnostic engine (that collects the conflicts and generate friendly diagnostics).

The top-level accumulation analyzer will still orchestrate these three engines, where the diagnostic engine acts like a conflict collector for both no-infer (`checkErrors`) and full-infer (`inference.Engine`) mode.

This PR also contains a few simple refactors and optimizations for the diagnostic engine for readability and performance.

Closes #55

### DIFF
--- a/accumulation/analyzer.go
+++ b/accumulation/analyzer.go
@@ -135,13 +135,14 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
 		// determining local and upstream sites in the process. This is guaranteed not to determine any
 		// sites unless we really have a reason they have to be determined.
 		inferenceEngine.ObservePackage(assertionsResult.FullTriggers)
-		inferredMap, diagnostics = inferenceEngine.InferredMap(), diagnosticEngine.Diagnostics(true /* grouping */)
+		inferredMap = inferenceEngine.InferredMap()
+		diagnostics = diagnosticEngine.Diagnostics(true /* grouping */)
 
 	case inference.NoInfer:
 		// In non-inference case - use the classical assertionNode.CheckErrors method to determine error outputs
 		inferredMap = inferenceEngine.InferredMap()
 		checkErrors(assertionsResult.FullTriggers, inferredMap, diagnosticEngine)
-		// Retrieve the diagnostics from the engine. Note that we do not need to group the
+		// Retrieve the diagnostics from the engine. Note that we should not group the
 		// diagnostics for easier unit testing.
 		diagnostics = diagnosticEngine.Diagnostics(false /* grouping */)
 
--- a/accumulation/analyzer.go
+++ b/accumulation/analyzer.go
@@ -26,6 +26,7 @@ import (
 	"go.uber.org/nilaway/assertion"
 	"go.uber.org/nilaway/assertion/function/assertiontree"
 	"go.uber.org/nilaway/config"
+	"go.uber.org/nilaway/diagnostic"
 	"go.uber.org/nilaway/inference"
 	"golang.org/x/tools/go/analysis"
 )
@@ -110,17 +111,19 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
 		return errorsToDiagnostics(errs), nil
 	}
 
+	diagnosticEngine := diagnostic.NewEngine(pass)
+
+	// Create an inference engine and observe (load) information from upstream dependencies (i.e.,
+	// mappings between annotation sites and their inferred values).
+	inferenceEngine := inference.NewEngine(pass, diagnosticEngine)
+	inferenceEngine.ObserveUpstream()
+
 	// Determine inference type based on comments in package doc string.
 	mode := inference.DetermineMode(pass)
 
-	// Create an engine and observe (load) information from upstream dependencies (i.e., mappings
-	// between annotation sites and their inferred values).
-	engine := inference.NewEngine(pass)
-	engine.ObserveUpstream()
-
 	// First observe all annotations from annotationsResult (observes only syntactic annotations
 	// for FullInfer mode, otherwise all annotations for NoInfer)
-	engine.ObserveAnnotations(annotationsResult.AnnotationMap, mode)
+	inferenceEngine.ObserveAnnotations(annotationsResult.AnnotationMap, mode)
 
 	var (
 		inferredMap *inference.InferredMap
@@ -131,13 +134,16 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
 		// Incorporate assertions from this package one-by-one into the inferredAnnotationMap, possibly
 		// determining local and upstream sites in the process. This is guaranteed not to determine any
 		// sites unless we really have a reason they have to be determined.
-		engine.ObservePackage(assertionsResult.FullTriggers)
-		inferredMap, diagnostics = engine.InferredMapWithDiagnostics()
+		inferenceEngine.ObservePackage(assertionsResult.FullTriggers)
+		inferredMap, diagnostics = inferenceEngine.InferredMap(), diagnosticEngine.Diagnostics(true /* grouping */)
 
 	case inference.NoInfer:
 		// In non-inference case - use the classical assertionNode.CheckErrors method to determine error outputs
-		inferredMap, diagnostics = engine.InferredMapWithDiagnostics()
-		diagnostics = append(diagnostics, buildDiagnostics(pass, checkErrors(assertionsResult.FullTriggers, inferredMap))...)
+		inferredMap = inferenceEngine.InferredMap()
+		checkErrors(assertionsResult.FullTriggers, inferredMap, diagnosticEngine)
+		// Retrieve the diagnostics from the engine. Note that we do not need to group the
+		// diagnostics for easier unit testing.
+		diagnostics = diagnosticEngine.Diagnostics(false /* grouping */)
 
 	default:
 		panic("Invalid mode for running NilAway")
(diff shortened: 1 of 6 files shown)


inference/engine.go
```diff
@@ -26,38 +26,43 @@ import (
...
+		diagnosticEngine: diagnosticEngine,
 	}
 }
 
-// InferredMapWithDiagnostics returns the current inferred annotation map and a slice of diagnostics
-// generated during inference.
-func (e *Engine) InferredMapWithDiagnostics() (*InferredMap, []analysis.Diagnostic) {
-	return e.inferredMap, e.conflicts.Diagnostics()
+// InferredMap returns the current inferred annotation map.
+func (e *Engine) InferredMap() *InferredMap {
```
yuxincs, line 64: This separates diagnostic generation logic from inference, callers should call `.Diagnostics()` on the diagnostic engine instead.

inference/explained_bool.go
```diff
@@ -28,10 +28,12 @@ import (
...
 type ExplainedBool interface {
+	fmt.Stringer
+
 	Val() bool
-	String() string
-	getPrimitiveFullTrigger() primitiveFullTrigger
-	deeperReason() ExplainedBool
+	Pos() token.Pos
+	TriggerReprs() (producer fmt.Stringer, consumer fmt.Stringer)
+	DeeperReason() ExplainedBool
```
yuxincs, line 36: We have to export these methods since diagnostic handling is in a separate package now. Moreover, it's often bad to have an exported method that returns an unexported struct. 

Since I think we don't want to export `primitiveFullTrigger` (it is only meaningful inside the inference engine), I have exposed these methods instead that are supposed to work across packages.

diagnostic/engine.go
```diff
@@ -144,112 +236,6 @@ func (c *conflict) addSimilarConflict(conflict conflict) {
 	c.similarConflicts = append(c.similarConflicts, &conflict)
 }
 
-// ConflictList stores a list of conflicts.
-type ConflictList struct {
-	conflicts  []conflict
-	NoGrouping bool // if set to true, conflicts are not grouped by nil path. (Example use case: no-infer unit tests)
```
yuxincs, line 150: This is removed and put as a parameter to `Diagnostics` instead, since it only affects the final diagnostic generation logic, but not the other parts (conflict collection logic).

diagnostic/engine.go
```diff
@@ -144,112 +236,6 @@ func (c *conflict) addSimilarConflict(conflict conflict) {
...
-func (l *ConflictList) AddSingleAssertionConflict(pass *analysis.Pass, trigger annotation.FullTrigger) {
-	t := fullTriggerAsPrimitive(pass, trigger)
-	c := conflict{
-		pos:  t.Pos,
-		flow: nilFlow{},
-	}
-
-	c.flow.addNonNilPathNode(t.ProducerRepr, t.ConsumerRepr)
-
-	l.conflicts = append(l.conflicts, c)
```
yuxincs, line 163: This is updated a little bit since we do not really need to convert this trigger to primitive trigger to have the necessary information (i.e., the producer and consumer prestrings)

diagnostic/engine.go
```diff
@@ -144,112 +236,6 @@ func (c *conflict) addSimilarConflict(conflict conflict) {
...
-
-	// Build nil path by traversing the inference graph from `nilExplanation` part of the overconstraint failure.
-	// (Note that this traversal gives us a backward path from point of conflict to the source of nilability. Hence, we
-	// must take this into consideration while printing the flow, which is currently being handled in `addNilPathNode()`.)
-	var queue []ExplainedBool
-	queue = append(queue, nilExplanation)
-
-	for len(queue) > 0 {
-		e := queue[0]
-		queue = queue[1:]
```
yuxincs, line 178: This logic is removed since we will only have one element in the queue at a time, meaning we do not really need to maintain a queue (having a moving ptr is good enough)

diagnostic/engine.go
```diff
@@ -144,112 +236,6 @@ func (c *conflict) addSimilarConflict(conflict conflict) {
...
-	// (Note that this traversal is forward from the point of conflict to dereference. Hence, we don't need to make
-	// any special considerations while printing the flow.)
-	// Different from building the nil path above, here we also want to deduce the position where the error should be reported,
-	// i.e., the point of dereference where the nil panic would occur. In NilAway's context this is the last node
-	// in the non-nil path. Therefore, we keep updating `c.pos` until we reach the end of the non-nil path.
-	queue = make([]ExplainedBool, 0)
-	queue = append(queue, nonnilExplanation)
-	for len(queue) > 0 {
-		e := queue[0]
-		queue = queue[1:]
```
yuxincs, line 208: Same here.

diagnostic/engine.go
```diff
@@ -144,112 +236,6 @@ func (c *conflict) addSimilarConflict(conflict conflict) {
...
-		if b := e.deeperReason(); b != nil {
-			queue = append(queue, b)
-		}
-	}
-
-	l.conflicts = append(l.conflicts, c)
-}
-
-// Diagnostics returns a list of diagnostics for the conflicts in the list.
-func (l *ConflictList) Diagnostics() []analysis.Diagnostic {
```
yuxincs, line 234: This now takes a `grouping bool` parameter to indicate whether the generated diagnostics should be grouped or not.

accumulation/analyzer.go
```diff
@@ -167,21 +173,13 @@ func errorsToDiagnostics(errs []error) []analysis.Diagnostic {
...
-	// setting NoGrouping to true since the no-infer mode is used via unit tests currently, and we actually want
-	// separate errors for better testability
-	conflicts := inference.ConflictList{NoGrouping: true}
-	for _, trigger := range triggers {
-		conflicts.AddSingleAssertionConflict(pass, trigger)
-	}
-	return conflicts.Diagnostics()
+type conflictHandler interface {
+	AddSingleAssertionConflict(trigger annotation.FullTrigger)
 }
```
yuxincs, line 178: Logic incorporated with `checkErrors` below

diagnostic/engine.go
```diff
@@ -12,18 +12,113 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package inference
+// Package diagnostic hosts the diagnostic engine, which is responsible for collecting the
+// conflicts from annotation-based checks (no-infer mode) and/or inference (full-infer mode) and
+// generating user-friendly diagnostics from those conflicts.
+package diagnostic
```
yuxincs, line 18: This file is moved from `inference/conflict.go`. Please see `inference/conflict.go` for the updates to the file content.

lazaroclapp (commented):
Overall, makes sense to me, some comments below!

diagnostic/engine.go
```diff
@@ -12,18 +12,113 @@
...
+	conflicts []conflict
+}
+
+// NewEngine creates a new diagnostic engine.
+func NewEngine(pass *analysis.Pass) *Engine {
+	return &Engine{pass: pass}
+}
+
+// Diagnostics generates diagnostics from the internally-stored conflicts. The grouping parameter
+// controls whether the conflicts with the same nil path are grouped together for concise reporting.
```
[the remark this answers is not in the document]
sonalmahajan15: Complete nil flow contains the entire path from a nilable source to a dereference point. Nil path here is referring to the part in the complete nil flow going from a nilable source point to the conflict point (refer [field nilPath in nilFlow](https://github.com/uber-go/nilaway/blob/fad4774c6a2e4747b238382cd718ed67f8f02466/diagnostic/engine.go#L129)). The grouping happens based on this `nilPath`.

Perhaps the comment can be rephrased to make it clearer.

inference/explained_bool.go
sonalmahajan15: Why not make the type of `producer` and `consumer` as `annotation.Prestring`? Any particular reason we want to use `fmt.Stringer`?

sonalmahajan15 (commented):
Overall, looks good to me. My only question is about the choice of using `fmt.Stringer` rather than `annotation.Prestring`.

diagnostic/engine.go
```diff
@@ -12,18 +12,113 @@
...
+			Message: c.String(),
+		})
+	}
+	return diagnostics
+}
+
+// AddSingleAssertionConflict adds a new single assertion conflict to the engine.
+func (e *Engine) AddSingleAssertionConflict(trigger annotation.FullTrigger) {
+	producer, consumer := trigger.Prestrings(e.pass)
+	flow := nilFlow{}
```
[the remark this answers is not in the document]
yuxincs: According to our [style guide](https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering), we should group functions by receivers (i.e., methods should follow the the type declarations), where the exported types and functions should appear earlier than unexported ones.

Actually, since we are now separating the diagnostic logic into a new package, we can just put the `conflict` and `nilFlow` type declarations into their own files for easier navigation.

Hopefully now it's clearer and easier to find them 😃

inference/explained_bool.go
yuxincs: `annotation.Prestring` is equivalent to `fmt.Stringer` in that they both expose `String() string` method and can be used interchangeably. So I opted for a standard `fmt.Stringer` instead of `annotation.Prestring`.

I was thinking we could remove `annotation.Prestring` interface definition from our codebase, it is already provided by stdlib and is equally readable (if not more). If we ever need to expand the interface, we should do it [at the client side](https://github.com/golang/go/wiki/CodeReviewComments#interfaces) instead of putting it in the `annotation` library package.

But I don't have strong preferences here and I can revert this to use `annotation.Prestring` if you disagree 😃

sonalmahajan15: I see your point. I agree that `fmt.Stringer` and `annotation.Prestring` are the essentially the same interfaces at this point. `fmt.Stringer` is also very less likely to change, since it is used quite ubiquitously, and if we want to expand the interface for our purpose, we can follow the path you are suggesting. Therefore, I am fine using either of them, provided we do so consistently across the entire codebase. 😄 You can keep this change, and perhaps create an issue to track changing of `Prestring` to `Stringer` at other places in the codebase.

sonalmahajan15 (approved):
LGTM! Only one follow-up task: to create an issue to track the change from `annotation.Prestring` to `fmt.Stringer` in other places in the codebase.

inference/explained_bool.go
yuxincs: 👍 !

I have created an issue #69 to track this clean-up task!

PR #63, opened 2023-09-28
Remove unused fields in primitive full trigger and unused utility functions

yuxincs:
This PR removes an unused field `ExprRepr` and an unused utility function for it. It was added when developing the new error message format. However, our design has evolved since then and this field is no longer needed. This should lead to a slight reduction of build time and artifact size overhead.

Depends on PR #62


lazaroclapp (approved):
Makes sense to me! This was used only for debugging-print representations of the triggers, right?

sonalmahajan15:
> Makes sense to me! This was used only for debugging-print representations of the triggers, right?

Not really. This was introduced in one of intermediate error message formats, but not needed anymore for the concise new message format.

sonalmahajan15 (approved):
LGTM!

PR #67, opened 2023-10-05
Use ordered map for inferred map

yuxincs:
In #66, we have created an ordered map data structure and used it for storing implication egdes.

This PR further refactors the `inference.InferredMap` to use the ordered map for deterministic evaluation of inferred maps.

This PR also adds another test for inferred map to check if the encoded binaries are the same over 10 iterations to catch nondeterminism in future development.

Depends on PR #66.


sonalmahajan15 (approved):
This PR by itself looks good to me.

PR #68, opened 2023-10-05
Sort upstream facts based on package paths

yuxincs:
We are calling `analysis.Pass.AllPackageFacts()` API from the analysis framework to import NilAway facts from upstream dependencies. However, this API is documented as

> AllPackageFacts returns a new slice containing all package facts of the analysis's FactTypes in **unspecified** order.

In some drivers such as nogo, this order is deterministic, however, in most other drives like the one from the `analysis` frameowork, the order will be nondeterministic.

Therefore, for determinism in NilAway, this PR sorts imported upstream facts (i.e., `inference.InferredMap`) based on the package path.

Depends on PR #67


go.mod
```diff
@@ -15,6 +15,6 @@ require (
 	github.com/kr/text v0.2.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	golang.org/x/mod v0.12.0 // indirect
-	golang.org/x/sys v0.10.0 // indirect
+	golang.org/x/sys v0.12.0 // indirect
```
sonalmahajan15, line 18: Is this and the below change to `go.sum` related to this PR?

sonalmahajan15 (commented):
LGTM, modulo the question asked below.

go.mod
yuxincs: Sorry this somehow got into this, it's a leftover from an experiment. Reverted 😃

sonalmahajan15 (approved):
LGTM!

PR #71, opened 2023-10-06
Fix panic in handling global variables

yuxincs:
This PR fixes two panics due to the assumption that the RHS of the global variable assignments are always variables, which isn't true when RHS is a constant (either defined in the current package, or referenced from upstream package)

We also take this opportunity to refactor the code a little bit to be more readable and maintainable.

Depends on #68

### DIFF
--- a/testdata/src/go.uber.org/globalvars/globalvarinit.go
+++ b/testdata/src/go.uber.org/globalvars/globalvarinit.go
@@ -19,6 +19,8 @@ These tests check if the nonnil global variables are initialized
 */
 package globalvars
 
+import "go.uber.org/globalvars/upstream"
+
 var x = 3
 
 // This should throw an error since it is not initialized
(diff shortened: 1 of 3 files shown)


testdata/src/go.uber.org/globalvars/globalvarinit.go
```diff
@@ -94,3 +96,14 @@ func foo() *int {
 	print(multiNonNil, multiNil, nonnilMethod, assignedNilable)
 	return nil
 }
+
+// Now we test a corner case where a constant is assigned to a global variable.
```
sonalmahajan15, line 100:
```suggestion
// Below test checks when a constant is assigned to a global variable.
```

sonalmahajan15: Feel free to ignore this suggestion. 🙂  My reasoning was that the comment with "Now" makes sense in the context of this PR, but the story might be lost later in code.

sonalmahajan15 (approved):
LGTM! Just a minor suggestion, feel free to ignore :)

Btw curious to know what was the panic here?

testdata/src/go.uber.org/globalvars/globalvarinit.go
yuxincs: I agree, good catch! updated!

PR #72, opened 2023-10-06
Fix panics when ranging over arrays

yuxincs:
This PR fixes a case where we are not unwrapping the pointer when checking if the range operand is deeply an array. This is causing panics since we have a safeguard on unknown types being ranged over, for example:

```
backpropagation across node (/opt/homebrew/Cellar/go/1.21.0/libexec/src/runtime/proc_test.go:599:6) of type *ast.AssignStmt failed for reason: unrecognized type of rhs in range statement: *[8192]byte
```

Depends on #71

### DIFF
--- a/testdata/src/go.uber.org/arrays/arrays.go
+++ b/testdata/src/go.uber.org/arrays/arrays.go
@@ -162,7 +162,7 @@ type blocks [42]int
 
 type blockSlice []int
 
-// nonnil(aPtr) nonnil(a) nonnil(bPtr) nonnil(b)
+// nonnil(aPtr, a, bPtr, b)
 func testArrayAliasPtr(aPtr *blocks, a blocks, bPtr *blockSlice, b blockSlice) {
 	// blocks is an alias for arrays, and the [language specs] states that it is possible to range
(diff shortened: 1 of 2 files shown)


testdata/src/go.uber.org/arrays/arrays.go
```diff
@@ -155,3 +155,32 @@ func test2dArrayAssignment() *int {
...
 	return twodArr[0][0]     //want "returned"
 }
+
+// Test a case where we declare a type alias for an array and then range over it.
+
+type blocks [42]int
+
+type blockSlice []int
+
+// nonnil(aPtr) nonnil(a) nonnil(bPtr) nonnil(b)
```
sonalmahajan15, line 165: Should work fine even with a single `nonnil()`

`// nonnil(aPtr, a, bPtr, b)`

sonalmahajan15 (approved):
LGTM!

testdata/src/go.uber.org/arrays/arrays.go
yuxincs: I was experimenting so I separated the annotations for quick edits. 

But yes, should've cleaned that up. Thanks!

PR #73, opened 2023-10-06
Fix backprop panic when ranging over a generic slice

yuxincs:
This PR fixes a panic when the code is ranging over a generic slice due to our safeguard on unhandled types.

Since we do not have proper support on generics yet, we are treating all generics slices as always being deeply nonnil. 

Depends on #72


sonalmahajan15 (approved):
Looks good

PR #78, opened 2023-10-12
Apply S2 compression when exporting the facts

yuxincs:
NilAway is writing a lot of data when exporting `inference.InferredMap` via the [Fact mechanism](https://pkg.go.dev/golang.org/x/tools/go/analysis#hdr-Modular_analysis_with_Facts) since we are storing information about the nilability values / chains about both exported / unexported sites.

While we try our best to limit the total amount of data written (e.g., storing compact structs in Facts and restore only when we need to report errors), we are still seeing ~3% artifact size overhead in our internal evaluation on a large repository.

This PR applies a fast and effective compression algorithm S2 when we serialize the inferred map via Gob Encoding. Our internal evaluation shows that the artifact size overhead decreases from ~3% down to ~0.4%, while the build time overhead increases from ~2% to ~2.9%. This seems a good compromise to make for future development, especially for enabling struct initialization support, where we expect a lot more triggers to be generated and hence more implication edges to be stored.

Depends on #73.

### DIFF
--- a/inference/inferred_map_test.go
+++ b/inference/inferred_map_test.go
@@ -34,9 +34,9 @@ func TestEncoding_Size(t *testing.T) {
 
 	out := buf.Bytes()
 	require.NotEmpty(t, out)
-	require.Less(t, len(out), 250_000,
+	require.Less(t, len(out), 30_000,
 		"The gob encoding of a test inferred map is too large. We expect the encoded "+
-			"map to be less than 250KB. This heavily affects the artifact sizes of the facts NilAway "+
+			"map to be less than 30KB. This heavily affects the artifact sizes of the facts NilAway "+
 			"produces, so the cap should only be increased with justification and thorough testing.",
 	)
 }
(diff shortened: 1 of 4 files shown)


inference/inferred_map_test.go
```diff
@@ -34,9 +34,9 @@ func TestEncoding_Size(t *testing.T) {
 
 	out := buf.Bytes()
 	require.NotEmpty(t, out)
-	require.Less(t, len(out), 250_000,
+	require.Less(t, len(out), 30_000,
 		"The gob encoding of a test inferred map is too large. We expect the encoded "+
-			"map to be less than 250KB. This heavily affects the artifact sizes of the facts NilAway "+
+			"map to be less than 30KB. This heavily affects the artifact sizes of the facts NilAway "+
```
yuxincs, line 39: Our current implementation w/ S2 compression yields ~27.3KB, so this bound is re-set to 30KB.

sonalmahajan15: Nice!

sonalmahajan15 (approved):
A much needed optimization! We have verified this internally, so everything looks great! :)

PR #81, opened 2023-10-12
Delete `String()` methods

sonalmahajan15:
This PR deletes `String()` methods that we no longer use.


yuxincs (approved):
Dead code deletion for easier maintenance, nice!

PR #82, opened 2023-10-12
Add `NeedsGuard()` in consumers

sonalmahajan15:
Similar to producers, this PR adds `NeedsGuard()` method to consumers, along with the `SetNeedsGuard()` method and `guard` field that can be updated. The PR also adds an explicit condition in `CheckGuardOnFullTrigger()` to check that the consumer needs a guard.

### DIFF
--- a/annotation/produce_trigger.go
+++ b/annotation/produce_trigger.go
@@ -50,7 +50,7 @@ type ProducingAnnotationTrigger interface {
 
 	// SetNeedsGuard sets the underlying Guard-Neediness of this ProduceTrigger, if present
 	// This should be very sparingly used, and only with utter conviction of correctness.
-	// Default setting for ProduceTriggers is to not need a guard. Override this method to set the need for a guard to true.
+	// Default setting for ProduceTriggers is to not need a guard.
 	SetNeedsGuard(bool)
 
 	Prestring() Prestring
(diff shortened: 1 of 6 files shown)


annotation/produce_trigger.go
```diff
@@ -49,7 +49,8 @@ type ProducingAnnotationTrigger interface {
 	NeedsGuardMatch() bool
 
 	// SetNeedsGuard sets the underlying Guard-Neediness of this ProduceTrigger, if present
-	// This should be very sparingly used, and only with utter conviction of correctness
+	// This should be very sparingly used, and only with utter conviction of correctness.
+	// Default setting for ProduceTriggers is to not need a guard. Override this method to set the need for a guard to true.
```
yuxincs, line 53:
```suggestion
	// Default setting for ProduceTriggers is to not need a guard. 
```

Since this is the interface definition, I think we probably shouldn't include implementation details. Also, we don't really need to override it, right? the base structs (`TriggerIfNilable` and `TriggerIfDeepNilable`) already properly implement this.

yuxincs (approved):
Logic LGTM modulo a documentation suggestion :)

annotation/produce_trigger.go
sonalmahajan15: Fair point.

PR #87, opened 2023-10-19
Refine backprop assignments in error message

### LINKED ISSUE #83: Refine error message with assignments
For code example below,
```
func foo(x *int) {
	x = nil
	y := x
	z := y
	print(*z) // L30. Error reported here
}
```

NilAway prints the error message as
```
Potential nil panic detected. Observed nil flow from source to dereference point: 
        	-> errormessage/errormessage.go:30:9: literal `nil` dereferenced
```

This can be confusing and difficult for the users to comprehend if the assignments have followed a complicated path. Instead, we would like NilAway to generate a more informative error message following the assignments.

sonalmahajan15:
This PR refines the error message by incorporating information about the assignments. Following is an example of an informative error message:

Code snippet:
```
func test(x *int) {
	x = nil
	y := x
	z := y
	print(*z)
}
```

Error message:
```
Potential nil panic detected. Observed nil flow from source to dereference point: 
   -> errormessage/errormessage.go:32:9: literal `nil` dereferenced via the assignment(s):
        -> `nil` to `x` at errormessage/errormessage.go:29:2,
        -> `x` to `y` at errormessage/errormessage.go:30:2,
        -> `y` to `z` at errormessage/errormessage.go:31:2
```

[closes #83 ]
[depends on #86 ]

### DIFF
--- a/annotation/consume_trigger.go
+++ b/annotation/consume_trigger.go
@@ -174,12 +174,6 @@ func (*ConsumeTriggerTautology) equals(other ConsumingAnnotationTrigger) bool {
 	return ok
 }
 
-// Copy returns a deep copy of this ConsumingAnnotationTrigger
-func (t *ConsumeTriggerTautology) Copy() ConsumingAnnotationTrigger {
-	copyConsumer := *t
-	return &copyConsumer
-}
-
 // Prestring returns this Prestring as a Prestring
 func (*ConsumeTriggerTautology) Prestring() Prestring {
 	return ConsumeTriggerTautologyPrestring{}
--- a/annotation/consume_trigger.go
+++ b/annotation/consume_trigger.go
@@ -19,6 +19,7 @@ import (
 	"go/ast"
 	"go/token"
 	"go/types"
+	"slices"
 	"strings"
 
 	"go.uber.org/nilaway/util"
@@ -53,6 +54,12 @@ type ConsumingAnnotationTrigger interface {
 
 	// Copy returns a deep copy of this ConsumingAnnotationTrigger
 	Copy() ConsumingAnnotationTrigger
+
+	// AddAssignment adds an assignment to the trigger for tracking and printing informative error message.
+	// NilAway's `backpropAcrossOneToOneAssignment()` lifts consumer triggers from the RHS of an assignment to the LHS.
+	// This implies loss of information about the assignment. This method is used to track such assignments and print
+	// a more informative error message.
+	AddAssignment(Assignment)
 }
 
 // customPos has the below default implementations, in which case ConsumeTrigger.Pos() will return a default value.
@@ -68,9 +75,61 @@ type Prestring interface {
 	String() string
 }
 
+// Assignment is a struct that represents an assignment to an expression
+type Assignment struct {
+	LhsExprStr string
+	RhsExprStr string
+	Position   token.Position
+}
+
+func (a *Assignment) String() string {
+	return fmt.Sprintf("`%s` to `%s` at %s", a.RhsExprStr, a.LhsExprStr, a.Position)
+}
+
+// assignmentFlow is a struct that represents a flow of assignments
+type assignmentFlow struct {
+	assignments []Assignment
+}
+
+func (a *assignmentFlow) addEntry(entry Assignment) {
+	a.assignments = append(a.assignments, entry)
+}
+
+func (a *assignmentFlow) String() string {
+	if len(a.assignments) == 0 {
+		return ""
+	}
+
+	// fix point convergence in backprop could mean duplicate entries in this slice, hence we filter out the duplicates here
+	entries := make([]Assignment, 0, len(a.assignments))
+	seen := make(map[string]bool)
+	for _, entry := range a.assignments {
+		if !seen[entry.String()] {
+			seen[entry.String()] = true
+			entries = append(entries, entry)
+		}
+	}
+
+	// backprop algorithm populates assignment entries in backward order. Reverse entries to get forward order of assignments.
+	slices.Reverse(entries)
+
+	// build string slice
+	strs := make([]string, len(entries))
+	for i, entry := range entries {
+		strs[i] = entry.String()
+	}
+
+	// build the informative print string tracking the assignments
+	var sb strings.Builder
+	sb.WriteString(" via the assignment(s):\n\t\t-> ")
+	sb.WriteString(strings.Join(strs, ",\n\t\t-> "))
+	return sb.String()
+}
+
 // TriggerIfNonNil is triggered if the contained Annotation is non-nil
 type TriggerIfNonNil struct {
 	Ann Key
+	assignmentFlow
 }
 
 // Kind returns Conditional.
@@ -100,21 +159,34 @@ func (t *TriggerIfNonNil) Copy() ConsumingAnnotationTrigger {
 	return &copyConsumer
 }
 
+// AddAssignment adds an assignment to the trigger.
+func (t *TriggerIfNonNil) AddAssignment(e Assignment) {
+	t.assignmentFlow.addEntry(e)
+}
+
 // Prestring returns this Prestring as a Prestring
-func (*TriggerIfNonNil) Prestring() Prestring {
-	return TriggerIfNonNilPrestring{}
+func (t *TriggerIfNonNil) Prestring() Prestring {
+	return TriggerIfNonNilPrestring{
+		AssignmentStr: t.assignmentFlow.String(),
+	}
 }
 
 // TriggerIfNonNilPrestring is a Prestring storing the needed information to compactly encode a TriggerIfNonNil
-type TriggerIfNonNilPrestring struct{}
+type TriggerIfNonNilPrestring struct {
+	AssignmentStr string
+}
 
-func (TriggerIfNonNilPrestring) String() string {
-	return "nonnil value"
+func (t TriggerIfNonNilPrestring) String() string {
+	var sb strings.Builder
+	sb.WriteString("nonnil value")
+	sb.WriteString(t.AssignmentStr)
+	return sb.String()
 }
 
 // TriggerIfDeepNonNil is triggered if the contained Annotation is deeply non-nil
 type TriggerIfDeepNonNil struct {
 	Ann Key
+	assignmentFlow
 }
 
 // Kind returns DeepConditional.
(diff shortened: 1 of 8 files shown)


util/util.go
```diff
@@ -491,3 +493,13 @@ func truncatePosition(position token.Position) token.Position {
...
+
+// ExprToString converts AST expression to string
+func ExprToString(e ast.Expr, pass *analysis.Pass) string {
+	var buf bytes.Buffer
+	err := printer.Fprint(&buf, pass.Fset, e)
+	if err != nil {
+		panic(fmt.Sprintf("Failed to convert AST expression to string: %v\n", err))
+	}
+	return buf.String()
+}
```
yuxincs, line 505: I'm having dejavu of this, didn't we add this in some other PRs?

Anyways, not something to fix :)

assertion/function/assertiontree/root_assertion_node.go
```diff
@@ -888,8 +888,6 @@ func getFuncLitFromAssignment(ident *ast.Ident) *ast.FuncLit {
 //	}
 //
 // ```
-//
-// nilable(path, result 0)
```
yuxincs, line 892: 👍

yuxincs (commented):
Overall LGTM 😃 but let's make sure the artifact size does not increase too much since we are adding a good amount of strings.

Maybe test after PR #78 which hopefully will make this a non-issue. (I'll try to merge that stack of PR ASAP, currently it is pending final performance validations)

annotation/consume_trigger.go
```diff
@@ -1562,13 +1836,18 @@ type UseAsErrorRetWithNilabilityUnknownPrestring struct {
...
+	var sb strings.Builder
 	if u.IsNamedReturn {
-		return fmt.Sprintf("found in at least one path of `%s()` for named return `%s` in position %d", u.FuncName, u.RetName, u.RetNum)
+		sb.WriteString(fmt.Sprintf("found in at least one path of `%s()` for named return `%s` in position %d", u.FuncName, u.RetName, u.RetNum))
+	} else {
+		sb.WriteString(fmt.Sprintf("found in at least one path of `%s()` for return in position %d", u.FuncName, u.RetNum))
 	}
-	return fmt.Sprintf("found in at least one path of `%s()` for return in position %d", u.FuncName, u.RetNum)
+	sb.WriteString(u.AssignmentStr)
+	return sb.String()
```
yuxincs, line 1850: Having read through this all, I'm actually a bit tempted to create a shared struct, say `metaPrestring` or something similar that stores `AssignmentStr` since it is shared by all of the prestring nodes. Then we just embed this struct in all Prestring structs.

Then, this `metaPrestring` can expose two methods, `writePrefix(io.Writer)` and `writeSuffix(io.Writer)`, which is called at the beginning and at the end of each `String()` function (by `u.meta.writePrefix(sb)`, since `strings.Builder` implements the `io.Writer` interface).

This would allow future expansion a little easier. 

But I also see that might be an overkill for now, so we can keep this in mind until we need to add another field to all prestrings :)

annotation/consume_trigger.go
```diff
@@ -71,9 +77,63 @@ type Prestring interface {
...
+	// fix point convergence in backprop could mean duplicate entries in this slice, hence we filter out the duplicates here
+	entries := make([]Assignment, 0, len(a.assignments))
+	seen := make(map[string]bool)
+	for _, entry := range a.assignments {
+		if !seen[entry.String()] {
+			seen[entry.String()] = true
+			entries = append(entries, entry)
+		}
+	}
+
```
yuxincs, line 114: Maybe we can use our own `orderedmap` implementation after PR #66 is merged (I'll try to merge that ASAP) such that we do not have to store a lot of entries and then dedup (i.e., we dedup on the fly).

But taking a step back, would this cause a lot of performance penalty? are there any other ways (maybe not) 🤔

assertion/function/assertiontree/backprop.go
```diff
@@ -578,7 +578,18 @@ buildShadowMask:
 					}
 
 					lhsNode, ok := rootNode.LiftFromPath(lpath)
-					if ok {
+					// TODO: below check for `lhsNode != nil` should not be needed when NilAway supports Ok form for
+					//  used-defined functions (tracked issue #77)
```
yuxincs, line 582: 👍 

(BTW this function is long enough to the point it's a bit painful to (re-)read every time there is a change to it, we probably want to refactor it to smaller pieces in future revisions).

testdata/src/go.uber.org/errormessage/errormessage.go
```diff
@@ -0,0 +1,96 @@
...
+
+var dummy bool
+
+func test1(x *int) {
+	x = nil
+	print(*x) //want "`nil` to `x`"
+}
+
+func test2(x *int) {
+	x = nil
```
yuxincs, line 29: Can we also add cases where `x` is 

(1) read from a map without guarding (default nilable)
(2) reading from a nil channel perhaps?
(3) an empty slice

assertion/function/assertiontree/backprop.go
```diff
@@ -602,10 +613,26 @@ buildShadowMask:
...
+
 						rootNode.AddProduction(&annotation.ProduceTrigger{
 							Annotation: rproducers[0].GetShallow().Annotation,
 							Expr:       lhsVal,
 						}, rproducers[0].GetDeepSlice()...)
+
+						afterLastIndex := len(rootNode.triggers)
+
+						// Update consumers of newly added triggers with assignment entries for informative printing of errors
+						if len(rootNode.triggers) > beforeLastIndex && len(rootNode.triggers) <= afterLastIndex {
```
yuxincs, line 627: Do we need this check?

(1) will `len(rootNode.triggers) > beforeLastIndex` ever be `false`? i.e., will `rootNode.triggers` actually shrink?

(2) `len(rootNode.triggers) == afterLastIndex` here, right?

annotation/consume_trigger.go
```diff
@@ -71,9 +77,63 @@ type Prestring interface {
 	String() string
 }
 
+// Assignment is a struct that represents an assignment to an expression
+type Assignment struct {
+	LHSExprStr string
+	RHSExprStr string
+	Position   token.Position
+}
```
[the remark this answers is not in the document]
sonalmahajan15: What you are suggesting logically makes sense, but there were a couple of reasons why I did not opt for this design choice.
1) `backpropAcrossOneToOneAssignment` which populates the `Assignment` struct object, does not have `*ast.AssignStmt`, but instead only has LHS/RHS `ast.Expr`s.
2) We can of course store `ast.Expr`s here. But since we want to convert the `ast.Expr` to string for printing, we'll additionally need to also store `*analysis.Pass` to facilitate that.

Hence, I thought exposing and storing only the bare minimum objects here might be more desirable. Let me know if you think otherwise. :)

util/util.go
sonalmahajan15: Haha, yes. This function was introduced in the revised error message format PR at one point, but removed later since it did not have any use. So, it was re-added in this PR since we want to convert LHS and RHS `ast.Expr` in an assignment statement to string.

annotation/consume_trigger.go
```diff
@@ -71,9 +77,63 @@ type Prestring interface {
...
+	var sb strings.Builder
+	sb.WriteString(" via the assignment(s):\n\t\t-> ")
+	sb.WriteString(strings.Join(strs, ",\n\t\t-> "))
+	return sb.String()
+}
+
 // TriggerIfNonNil is triggered if the contained Annotation is non-nil
 type TriggerIfNonNil struct {
 	Ann Key
+	assignmentFlow
```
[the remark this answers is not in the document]
sonalmahajan15: Now, with using `orderedmap` for `assignmentFlow`, we should consider it in `copy`. Added code for it.

For `equals` we should not account for it. Consider the situation of multiple nilable flows reach a dereference site. We keep NilAway practical by tracking only one representative nil flow, and printing only one error message for the dereference site with one representative flow printed in the error message. However, considering `assignmentFlow` in `equals` would mean tracking all nilable flows in separate full triggers and reporting errors through all paths. This would likely incur a performance penalty, which we wouldn't want. I have added a note at the definition of `assignmentFlow` explaining this.

annotation/consume_trigger.go
```diff
@@ -71,9 +77,63 @@ type Prestring interface {
...
+	// backprop algorithm populates assignment entries in backward order. Reverse entries to get forward order of assignments.
+	for i, j := 0, len(entries)-1; i < j; i, j = i+1, j-1 {
+		entries[i], entries[j] = entries[j], entries[i]
+	}
+
+	// build string slice
+	strs := make([]string, len(entries))
+	for i, entry := range entries {
+		strs[i] = entry.String()
+	}
```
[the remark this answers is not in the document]
sonalmahajan15: Good point. Done in the context of using `orderedmap`.

annotation/consume_trigger.go
sonalmahajan15: Totally agreed! Every time I visit `consume_trigger.go` and `produce_trigger.go`, I am tempted to rearchitect it for Prestrings. I feel there is a lot of repetition there. A problem to be solved some other day :)

annotation/consume_trigger.go
sonalmahajan15: Done. Using `orderedmap` now :)

sonalmahajan15: Haven't observed any performance penalty as such, but will run a thorough performance experiment to confirm.

assertion/function/assertiontree/backprop.go
sonalmahajan15: Oh yes, agreed. This can definitely benefit from refactoring :)

testdata/src/go.uber.org/errormessage/errormessage.go
sonalmahajan15: Added.

assertion/function/assertiontree/backprop.go
sonalmahajan15: You are right, revised the checks.

annotation/consume_trigger.go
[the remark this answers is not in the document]
yuxincs: I see!

The only concern I have is the memory consumption and artifact size increases (we are already getting issue reports stating that NilAway consumes memory a lot). Let's 

(1) run `make build && bin/nilaway -memprofile ./mem.prof std`, then `go tool pprof -alloc_space ./mem.pprof` and then type `top`, which should show the sum of allocated memory. This should be a good proxy to see the memory consumption. We could simply compare before & after this PR to see if it has a huge impact on memory.

(2) we should run performance validations internally to see if it has an impact on artifact size.

Other than that, this makes sense to me 👍

yuxincs (approved):
Logic LGTM! The only concern I have is memory consumption & artifact size increases.

Let's do a quick profiling & internal performance validations to ensure the consumptions are still within reasonable bounds 😃

annotation/consume_trigger.go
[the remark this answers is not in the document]
sonalmahajan15: Fair point! I performed both the suggestions above: memory profiling and internal performance evaluation. There was a slight increase in memory use, but well within the acceptable bounds. The performance evaluation did not show any noticeable change. In summary, I have confirmed that the changes in this PR do not incur any significant overhead.

sonalmahajan15:
> Logic LGTM! The only concern I have is memory consumption & artifact size increases.
> 
> Let's do a quick profiling & internal performance validations to ensure the consumptions are still within reasonable bounds 😃

Checked with profiling and internal performance validation. Both memory consumption and artifact size within acceptable bounds.

PR #89, opened 2023-11-01
Upgrade x/tools to v0.14.0

yuxincs:
This PR upgrades the dependency `x/tools` (containing the `analysis` framework) to v0.14.0 since it includes various optimizations for the linter driver framework. Most notably, it includes fact storing logic optimization (see this [commit](https://github.com/golang/tools/commit/d75c38746e7bc937ac6208cc0549fd192f0d4b48)) such that imported facts are not re-exported, which should bring some performance improvements, as well as artifact size reduction.


sonalmahajan15 (approved):
LGTM :)

mauri870:
@yuxincs x/tools v0.11 is causing issues on go tip and apparently 1.20.6 in some codebases, see https://github.com/uber-go/nilaway/issues/116.

This MR could be changed to use x/tools v0.15.0 while we are at it.

Thanks!

dolmen:
@mauri870 See #96.

This one can be closed.

PR #133, opened 2023-11-23
Add run Makefile target

ashmckenzie:
This PR adds a `run` Makefile target to make it easier to well, run the tool during development/testing 🙂

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,10 @@ all: build lint test
 build:
 	go install go.uber.org/nilaway/cmd/nilaway
 
+.PHONY: run
+run:
+	go run cmd/nilaway/main.go ./...
+
 .PHONY: test
 test:
 	go test -v -race ./...


yuxincs:
Hi @ashmckenzie , thanks for your contributions!

The reason we did not include `make run` initially is because you won't be able to pass arguments to the nilaway, which is usually important and useful (for error suppressions and/or tweaking configurations). We can use an env var to pass it like `ARG="-foo bar -bar baz" make run`, but we didn't really pursue that since you can always just `make build && bin/nilaway -foo bar -bar baz ...`

Would you mind elaborating your use cases?

ashmckenzie:
That's fair @yuxincs, we also find passing args/options to `make` in our various GitLab projects to be awkward too 😞 

I found running `make run` valuable when I was attempting to add a 'summary line' to the end of the run to make sure things still compiled, but there are other ways to achieve this, so let's close it.

Thanks for your time @yuxincs 🙇

PR #140, opened 2023-11-28
Use objectpath.Encoder to amortize the cost of encoding the paths of multiple objects

yuxincs:
We are using `objectpath.For` very heavily in our inference engine to construct package-independent object identifiers. However, each invocation comes with a cost of initializations (construction and traversal of the scope tree). 

Since x/tools v0.9.0 (we are currently on v0.11.0), there is an `objectpath.Encoder` that "[amortizes the cost of encoding the paths of multiple objects](https://pkg.go.dev/golang.org/x/tools/go/types/objectpath#Encoder)". This PR uses the encoder for better performance. There aren't/shouldn't be any functionality changes.

A quick profile showing the timing before & after this PR for `accumulation.run` function (the entrypoint function for our inference engine):

before:

```
31.43s 22.49% |   go.uber.org/nilaway/accumulation.run
15.69s 49.92% |   go.uber.org/nilaway/inference.(*Engine).ObservePackage
8.11s 25.80%  |   go.uber.org/nilaway/inference.(*Engine).ObserveUpstream
3.33s 10.59%  |   go.uber.org/nilaway/inference.(*InferredMap).Export
...
```

after:

```
18.34s 14.63% |   go.uber.org/nilaway/accumulation.run
8.55s 46.62%  |   go.uber.org/nilaway/inference.(*Engine).ObserveUpstream
3.65s 19.90%  |   go.uber.org/nilaway/inference.(*InferredMap).Export
2.50s 13.63%  |   go.uber.org/nilaway/inference.(*Engine).ObservePackage
...
```

### DIFF
--- a/inference/primitive.go
+++ b/inference/primitive.go
@@ -125,6 +125,9 @@ type primitivizer struct {
 	// curDir is the current working directory, which is used to trim the prefix (e.g.,  random
 	// sandbox prefix if using bazel) from the file names for cross-package references.
 	curDir string
+	// objPathEncoder is used to encode object paths, which amortizes the cost of encoding the
+	// paths of multiple objects.
+	objPathEncoder *objectpath.Encoder
 }
 
 // newPrimitivizer returns a new and properly-initialized primitivizer.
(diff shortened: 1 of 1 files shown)


sonalmahajan15 (approved):
Wow, this is quite promising! Let's confirm the benefits via performance evaluation internally, as well as, a sanity check for ensuring there are no functionality changes, i.e., the errors reported for a given set of test targets remain the same.

PR #41, opened 2023-08-18
Support general nonnil->nonnil function contract, i.e., having a single nonnil but any numbers of any, e.g. contract(_,nonnil->nonnil,_)

zzqatuber:
Support general nonnil->nonnil function contract, i.e., having a single nonnil but any numbers of any, e.g. contract(_,nonnil->nonnil,_).

PR #141, opened 2023-11-28
Enable manual contract support in NilAway

yuxincs:
@zzqatuber has implemented contract support in NilAway and has merged the manual contract support (i.e., via manual annotations such as `//contract(nonnil->nonnil)`) in main NilAway but hidden under a flag.

After internal performance validations there are no major performance/functionality degradations of simply enabling this feature, therefore this PR removes the hidden flag and make it available in NilAway.

After this PR, we will prioritize merging the automatic inference of the function contracts (i.e., PR #40 , PR #41 , and PR #42 ).

### DIFF
--- a/annotation/map.go
+++ b/annotation/map.go
@@ -627,7 +627,7 @@ func newObservedMap(pass *analysis.Pass, files []*ast.File) *ObservedMap {
 
 	// Parse inline annotations at call sites.
 	for _, file := range files {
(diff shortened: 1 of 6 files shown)


sonalmahajan15 (approved):
First step in enabling the much needed contract support! :) These changes have been verified internally, so everything looks good to me.

PR #155, opened 2023-12-21
Fix error handling logic to analyze functions with mixed return types

sonalmahajan15:
The error handling logic contained code that skipped analysis of error returning functions with a mix of nilable and non-nilable (e.g., string, int) types, resulting in false negatives. This PR fixes the bug.

### DIFF
--- a/assertion/function/assertiontree/backprop_util.go
+++ b/assertion/function/assertiontree/backprop_util.go
@@ -280,13 +280,6 @@ func handleErrorReturns(rootNode *RootAssertionNode, retStmt *ast.ReturnStmt, re
 	nonErrRetExpr := results[:errRetIndex] // n-1 expressions
 
 	// check if the error return is at all guarding any nilable returns, such as pointers, maps, and slices
-	for _, r := range nonErrRetExpr {
-		if util.ExprBarsNilness(rootNode.Pass(), r) {
-			// no need to further analyze and create triggers
-			return true
-		}
-	}
-
 	if isErrorReturnNil(rootNode, errRetExpr) {
 		// if error is the only return expression in the statement, then create a consumer for it, else create consumers for the non-error return expressions
 		if len(nonErrRetExpr) == 0 {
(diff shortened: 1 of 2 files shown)


assertion/function/assertiontree/backprop_util.go
```diff
@@ -267,14 +267,6 @@ func handleErrorReturns(rootNode *RootAssertionNode, retStmt *ast.ReturnStmt, re
 	errRetExpr := results[errRetIndex]     // n-th expression
 	nonErrRetExpr := results[:errRetIndex] // n-1 expressions
 
-	// check if the error return is at all guarding any nilable returns, such as pointers, maps, and slices
-	for _, r := range nonErrRetExpr {
-		if util.ExprBarsNilness(rootNode.Pass(), r) {
-			// no need to further analyze and create triggers
-			return
```
sonalmahajan15, line 274: Safe to remove this logic, since `ExprBarsNilness` is checked before creating the consume triggers anyway.

testdata/src/go.uber.org/errorreturn/inference/errorreturn-with-inference.go
```diff
@@ -176,3 +176,37 @@ func checkAndDeref() {
 		_ = *x //want "error return in position 1 is not guaranteed to be non-nil through all paths"
 	}
 }
+
+// ***** below tests error handling logic for mixed nilable (e.g., pointer) and non-nilable (e.g., string) non-error returns *****
```
yuxincs, line 180: Nice! Could we add a case for aliased ptr type?

yuxincs (approved):
Nice catch! I only have a small suggestion for adding a test case (but not strong preference here)

testdata/src/go.uber.org/errorreturn/inference/errorreturn-with-inference.go
sonalmahajan15: Good idea. Added the test case  :)

PR #156, opened 2023-12-21
Delete unused `String()` methods

sonalmahajan15:
This PR performs code cleanup by deleting unused `String()` methods.

PR #157, opened 2023-12-21
Support for handling of `ok` form for functions/methods

### LINKED ISSUE #77: Support `ok` form for functions and methods
Add support for modeling `bool` returns and their corresponding `ok` forms. This support can be modeled similar to the current `error` return handling in NilAway.

```
func test() {
	m := map[*int]*int{}
	if v, ok := m[new(int)]; ok {
		print(*v) // no error reported because NilAway handles "OK" form for a map.
	}
	if v, ok := foo(new(int)); ok {
		print(*v) // However, this line still throws a false positive because NilAway does not handle "OK" form for a user-defined function.
	}
}

func foo(x *int) (*int, bool) {
	if x == nil {
		return nil, false
	}
	return new(int), true
}
```

sonalmahajan15:
This pull request introduces support for the `ok` form in both user-defined and library functions and methods. The implementation addresses false positives, such as those identified in issue #77.

Currently, the feature is designed to handle explicit boolean returns, specifically in the form of `return r0, r1, ..., true`. Support for expression-based returns (e.g., `return r0, r1, ..., flag` or `return r0, r1, ..., isOk()`) is tricky, as tracking boolean types is currently not supported in NilAway. We can handle this scenario in the future.

[Closes #77 ]
[Depends on #156 ]

### DIFF
--- a/assertion/function/assertiontree/backprop.go
+++ b/assertion/function/assertiontree/backprop.go
@@ -198,13 +198,12 @@ func backpropAcrossReturn(rootNode *RootAssertionNode, node *ast.ReturnStmt) err
 						// this nil check reflects programmer logic
 						return errors.New("producers variable is nil")
 					}
-
-					isErrReturning := util.FuncIsErrReturning(funcObj)
-					isOkReturning := util.FuncIsOkReturning(funcObj)
-
 					// since we don't individually track the returns of a multiply returning function,
 					// we form full triggers for each return whose type doesn't bar nilness
 					if !util.TypeBarsNilness(funcObj.Type().(*types.Signature).Results().At(i).Type()) {
+						isErrReturning := util.FuncIsErrReturning(funcObj)
+						isOkReturning := util.FuncIsOkReturning(funcObj)
+
 						rootNode.AddNewTriggers(annotation.FullTrigger{
 							Producer: &annotation.ProduceTrigger{
 								// since the value is being returned directly, only its shallow nilability
--- a/assertion/function/assertiontree/backprop.go
+++ b/assertion/function/assertiontree/backprop.go
@@ -200,6 +200,7 @@ func backpropAcrossReturn(rootNode *RootAssertionNode, node *ast.ReturnStmt) err
 					}
 
 					isErrReturning := util.FuncIsErrReturning(funcObj)
+					isOkReturning := util.FuncIsOkReturning(funcObj)
 
 					// since we don't individually track the returns of a multiply returning function,
 					// we form full triggers for each return whose type doesn't bar nilness
@@ -225,7 +226,7 @@ func backpropAcrossReturn(rootNode *RootAssertionNode, node *ast.ReturnStmt) err
 								// if an error returning function returns directly as the result of
 								// another error returning function, then its results can safely be
 								// interpreted as guarded
-								GuardMatched: isErrReturning,
+								GuardMatched: isErrReturning || isOkReturning,
 							},
 						})
 					}
--- a/assertion/function/assertiontree/backprop_util.go
+++ b/assertion/function/assertiontree/backprop_util.go
@@ -137,8 +137,7 @@ func computeAndConsumeResults(rootNode *RootAssertionNode, node *ast.ReturnStmt)
 			}
 
 			// if the function has named error return variable, then handle specially using the error handling logic
-			if util.FuncIsErrReturning(rootNode.FuncObj()) {
-				handleErrorReturns(rootNode, node, results, true /* isNamedReturn */)
+			if ok := handleErrorReturns(rootNode, node, results, true /* isNamedReturn */); ok {
 				return nil
 			}
 
@@ -205,8 +204,7 @@ func computeAndConsumeResults(rootNode *RootAssertionNode, node *ast.ReturnStmt)
 		)
 	}
 
-	if util.FuncIsErrReturning(rootNode.FuncObj()) {
-		handleErrorReturns(rootNode, node, node.Results, false /* isNamedReturn */)
+	if ok := handleErrorReturns(rootNode, node, node.Results, false /* isNamedReturn */); ok {
 		return nil
 	}
 	if ok := handleBooleanReturns(rootNode, node, node.Results, false /* isNamedReturn */); ok {
@@ -272,7 +270,11 @@ func isErrorReturnNonnil(rootNode *RootAssertionNode, errRet ast.Expr) bool {
 // (3) if error return value = unknown, create consumers for all returns (error and non-error), and defer applying of the error contract when the nilability status is known, such as at `ProcessEntry`
 //
 // Note that `results` should be explicitly passed since `retStmt` of a named return will contain no results
-func handleErrorReturns(rootNode *RootAssertionNode, retStmt *ast.ReturnStmt, results []ast.Expr, isNamedReturn bool) {
+func handleErrorReturns(rootNode *RootAssertionNode, retStmt *ast.ReturnStmt, results []ast.Expr, isNamedReturn bool) bool {
+	if !util.FuncIsErrReturning(rootNode.FuncObj()) {
+		return false
+	}
+
 	errRetIndex := len(results) - 1
 	errRetExpr := results[errRetIndex]     // n-th expression
 	nonErrRetExpr := results[:errRetIndex] // n-1 expressions
@@ -281,7 +283,7 @@ func handleErrorReturns(rootNode *RootAssertionNode, retStmt *ast.ReturnStmt, re
 	for _, r := range nonErrRetExpr {
 		if util.ExprBarsNilness(rootNode.Pass(), r) {
 			// no need to further analyze and create triggers
-			return
+			return true
 		}
 	}
 
@@ -314,6 +316,7 @@ func handleErrorReturns(rootNode *RootAssertionNode, retStmt *ast.ReturnStmt, re
 			}
 		}
 	}
+	return true
 }
 
 // handleBooleanReturns handles the special case for boolean (`ok`) returning functions (n-th result of type `bool`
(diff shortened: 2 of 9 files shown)


assertion/function/assertiontree/backprop_util.go
```diff
@@ -307,6 +319,33 @@ func handleErrorReturns(rootNode *RootAssertionNode, retStmt *ast.ReturnStmt, re
 	}
 }
 
+// handleBooleanReturns handles the special case for boolean (`ok`) returning functions (n-th result of type `bool`
+// which guards at least one of the first n-1 non-bool results). Similar to the handliong of error returning functions,
+// for boolean returns, we generate consumers by applying the following boolean contract:
+// (1) if boolean return value = true, create consumers for the non-boolean returns
+// TODO: currently we support only explicit boolean returns (i.e., `return r0, r1, ..., {true|false}`). We should also support implicit boolean returns, i.e., `return` or `return <expr>` in the future.
+//
+// handleBooleanReturns returns true if the above contract is satisfied and consumers are created, false otherwise
```
yuxincs, line 328: Can we add some notes in the comments stating that the length of `results` will be >= 2? I think this is guaranteed by `util.IsFuncXXX` check in the caller, right?

assertion/function/assertiontree/backprop_util.go
```diff
@@ -307,6 +319,33 @@ func handleErrorReturns(rootNode *RootAssertionNode, retStmt *ast.ReturnStmt, re
...
+func handleBooleanReturns(rootNode *RootAssertionNode, retStmt *ast.ReturnStmt, results []ast.Expr, isNamedReturn bool) bool {
+	nRetIndex := len(results) - 1
+	nRetExpr := results[nRetIndex]          // n-th expression
+	nMinusOneRetExpr := results[:nRetIndex] // n-1 expressions
+
+	// check if the return statement is of the currently supported explicit boolean return form (`return ..., {true|false}`)
+	val, ok := nRetExpr.(*ast.Ident)
+	if !ok || (val.Name != "true" && val.Name != "false") {
+		return false
+	}
```
yuxincs, line 338: I remember we have a `util.IsLiteral` helper function to check if an expression is a literal nil, true, or false. Right? Let's use that :) 

Another thought, for a follow-up PR, is to enhance the `IsLiteral` function: currently, we are checking if the AST node is an identifier and the name is either `true` or `false`. This is error-prone (although unlikely to happen in real production code): devs can 

(1) shadow `true` as a variable:
```go
func foo() bool {
  var true bool = false
  return true // whoops, nilaway thinks this is true
}
```

(2) return a compile-time constant:

```go
const aa = true

func foo() bool {
  return aa // nilaway doesn't think this is returning true.
}
```

We should leverage proper type-checking and use `pass.TypesInfo.Type[expr]` to get a https://pkg.go.dev/go/types#TypeAndValue, and then use https://pkg.go.dev/go/constant#Val to check its compile-time value. We could probably also put it in a helper function in, say `typeshelper` :)

assertion/function/assertiontree/contract.go
```diff
@@ -228,18 +232,18 @@ func parseExpr(rootNode *RootAssertionNode, expr ast.Expr) TrackableExpr {
...
 }
 
-// NodeTriggersOkRead is a case of a node creating a rich bool effect for map read and channel receive in the "ok" form.
-// It matches on `AssignStmt`s of the form `v, ok := mp[k]` and `v, ok := <-ch`
-// nilable(result 0)
+// NodeTriggersOkRead is a case of a node creating a rich bool effect for map reads, channel receives, and user-defined
+// functions in the "ok" form. Specifically, it matches on `AssignStmt`s of the form
+// - `v, ok := mp[k]`
+// - `v, ok := <-ch`
+// - `r0, r1, r2, ..., ok := f()`
```
yuxincs, line 239: Nice additions!

util/util.go
```diff
@@ -303,35 +303,59 @@ func IsEmptyExpr(expr ast.Expr) bool {
...
+// A function is deemed "rich-check-effect-returning" iff it has a single result of type `typName` (error or bool),
+// and that result is the last in the list of results.
+func funcIsRichCheckEffectReturning(fdecl *types.Func, typName string) bool {
 	results := fdecl.Type().(*types.Signature).Results()
 	n := results.Len()
 	if n == 0 {
 		return false
 	}
-	if !TypeIsErrorType(results.At(n - 1).Type()) {
+	if !TypeIsRichCheckEffectType(results.At(n-1).Type(), typName) {
```
yuxincs, line 332: It's safer to not check against strings. 

See https://github.com/uber-go/nilaway/blob/main/util/util.go#L31, we can look up the Universe scope to get the builtin types and then directly compare against that (speaking of which, we should probably move that to `util/typeshelper` as well).

This would prevent us (again, unlikely to happen in production code) from misunderstanding a shadowed type, e.g., nothing prevents devs to do

```go
type error int
```

(perhaps let's add a test case for that)

assertion/function/assertiontree/contract.go
```diff
@@ -299,6 +315,37 @@ func NodeTriggersOkRead(rootNode *RootAssertionNode, nonceGenerator *util.GuardN
...
 		}
+	case *ast.CallExpr:
+		callIdent := util.FuncIdentFromCallExpr(rhs)
+		if callIdent == nil {
+			// this discards the case of an anonymous function
+			// perhaps in the future we could change this
+			return nil, false
+		}
+
+		rhsFuncDecl, ok := rootNode.Pass().TypesInfo.ObjectOf(callIdent).(*types.Func)
```
yuxincs, line 326: Let's call `rootNode.ObjectOf`, which is a wrapper that can handle fake idents :)

testdata/src/go.uber.org/contracts/userdefinedfunctions.go
```diff
@@ -0,0 +1,733 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package contracts: This file tests the contract of "ok" form for user defined and standard library functions.
+//
+// <nilaway no inference>
+package contracts
```
yuxincs, line 18: Appreciate the comprehensive test suite. As mentioned in other comments, perhaps we can add some cases that devs shadow the builtin types, as well as defining a constant for `true` or `false` :)

yuxincs (commented):
Nice feature addition!

> Support for expression-based returns (e.g., return r0, r1, ..., flag or return r0, r1, ..., isOk()) is tricky, as tracking boolean types is currently not supported in NilAway

I was a bit skeptical at first thinking that returning a constant boolean is not very common, but it turns out I was wrong. The errors NilAway reports on stdlib reduced 46 (let's sample a few to make sure they make sense). 

Logic overall LGTM, but I have some suggestions on code simplification / handling against shadowing builtin names. :) not all have to be handled in this single PR of course.

assertion/function/assertiontree/backprop_util.go
```diff
@@ -307,6 +319,33 @@ func handleErrorReturns(rootNode *RootAssertionNode, retStmt *ast.ReturnStmt, re
 	}
 }
 
+// handleBooleanReturns handles the special case for boolean (`ok`) returning functions (n-th result of type `bool`
+// which guards at least one of the first n-1 non-bool results). Similar to the handliong of error returning functions,
```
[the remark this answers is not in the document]
sonalmahajan15: Done :)

assertion/function/assertiontree/backprop_util.go
```diff
@@ -140,6 +141,12 @@ func computeAndConsumeResults(rootNode *RootAssertionNode, node *ast.ReturnStmt)
 				return nil
 			}
 
+			if util.FuncIsOkReturning(rootNode.FuncObj()) {
+				if ok := handleBooleanReturns(rootNode, node, results, true /* isNamedReturn */); ok {
+					return nil
+				}
+			}
```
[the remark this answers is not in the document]
sonalmahajan15: That's a good idea. Done :)

assertion/function/assertiontree/backprop_util.go
sonalmahajan15: That's correct. Added a comment.

util/util.go
sonalmahajan15: Ah, that's an interesting pointer. Implemented this safeguard and added test cases under `contracts/namedtypes`. (To avoid cluttering this PR, I'll move to `util/typeshelper` in a follow-up PR.)

assertion/function/assertiontree/contract.go
sonalmahajan15: Done!

testdata/src/go.uber.org/contracts/userdefinedfunctions.go
sonalmahajan15: Done for constants. Shadowing built-in types is not so straightforward, so will do it subsequently.

util/util.go
```diff
@@ -303,35 +303,59 @@ func IsEmptyExpr(expr ast.Expr) bool {
...
+}
+
+// FuncIsOkReturning encodes the conditions that a function is deemed "ok-returning".
+// This guards its results to require an `ok` check before use as nonnil.
+// A function is deemed "ok-returning" iff it has a single result of type `bool`, and that
+// result is the last in the list of results.
+func FuncIsOkReturning(fdecl *types.Func) bool {
+	return funcIsRichCheckEffectReturning(fdecl, _boolType)
+}
+
```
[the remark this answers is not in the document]
sonalmahajan15: Agreed. I will do this movement in a follow-up PR to keep this PR clutter-free :)

assertion/function/assertiontree/backprop.go
```diff
@@ -200,6 +200,7 @@ func backpropAcrossReturn(rootNode *RootAssertionNode, node *ast.ReturnStmt) err
 					}
 
 					isErrReturning := util.FuncIsErrReturning(funcObj)
+					isOkReturning := util.FuncIsOkReturning(funcObj)
```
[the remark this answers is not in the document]
sonalmahajan15: Good point. Done :)

assertion/function/assertiontree/backprop_util.go
sonalmahajan15: You bring up an interesting point! Although rare, you are right that we should account for shadowed built-ins and compile time constants. (2) is done, since it was fairly straightforward to implement. (1) needs more work, which I'll try to get in this PR, or maybe a follow-up. However, the strengthening of the `isLiteral` will definitely be done in a follow-up, since it is used at other places unrelated to this PR as well.

yuxincs (approved):
Great work! I see my comments are all addressed (where the further refactors will happen in follow-up PRs), thanks!

PR #171, opened 2024-01-16
Handle more cases for zero-slicing nilable slices

### LINKED ISSUE #168: Handle more cases for zero-slicing nilable slices
We have some handlings for zero-slicing nilable slices (e.g., `nilableSlice[0:]`, `nilableSlice[:0]` etc.). However, the following code piece still triggers a NilAway error:

```
func foo() {
	var a []int
	a = nil
	a = append(a[:0:0], a[0:]...)
}
```
```
unexpected diagnostic: Potential nil panic detected. Observed nil flow from source to dereference point: 
        	-> consts/consts.go:17:13: literal `nil` sliced into
```

Proof that it is actually nilsafe: https://go.dev/play/p/YRtwmo1r_IM

We should investigate and add support for such cases.

matsuyoshi30:
Support more cases for zero-slicing nilable slices like `[:0:0]` (fix #168)

### DIFF
--- a/assertion/function/assertiontree/root_assertion_node.go
+++ b/assertion/function/assertiontree/root_assertion_node.go
@@ -779,7 +779,7 @@ func (r *RootAssertionNode) AddComputation(expr ast.Expr) {
 	case *ast.SliceExpr:
 		// similar to index case
 
-		// zero slicing contains b[:0] b[0:0] b[0:] b[:] b[0:0:0], which are safe even when b is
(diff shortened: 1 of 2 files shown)


sonalmahajan15 (commented):
Thanks for the PR! The code changes and the test cases look good to me. I have a couple of straightforward asks before approving the PR:
- add description for the PR
- update the comment at [root_assrtion_node.go:L782](https://github.com/matsuyoshi30/nilaway/blob/fix-168/assertion/function/assertiontree/root_assertion_node.go#L782) to add this new case of `[:0:0]`

sonalmahajan15 (approved):
LGTM :)

PR #176, opened 2024-01-22
Refactor the docContains functions to asthelper package

yuxincs:
This PR does a simple refactor that consolidates the group of `docContains` functions and move them to the new `asthelper` package for better code organization and maintainability. No functionality changes are included in this PR

PR #177, opened 2024-01-22
Simplify struct initialization feature flag

yuxincs:
We currently have different modes for struct initialization feature flag (no check vs depth one check). However, we've not been actively adding more modes to this feature. On the opposite, we are trying to merge struct initialization feature into main NilAway itself. 

This PR simplifies this feature flag to be a simple boolean (ON/OFF) to make it easier for further refactors (i.e., exposing this experimental feature as a top-level flag). We can always expand if we truly need the support in the future (which doesn't seem to be the case for now), and simplifying such a flag helps with maintainability and readability a lot.

Depends on PR #176


sonalmahajan15 (approved):
LGTM :)

PR #178, opened 2024-01-22
Expose struct initialization and anonymous function support as top-level flags

yuxincs:
We have struct initialization and anonymous function support hidden under magic file headers for testing purposes, while we try our best to optimize and merge those features into main NilAway. However, we've had a fair number of issue reports asking why NilAway behaves incorrectly, where the problem is already solved by the corresponding features.

This PR makes it easier for users to supply a feature flag to the top-level driver (i.e., `-experimental-struct-init` and `-experimental-anonymous-function`) to enable such features.

Moreover, this PR simplifies the process by turning ON/OFF such a feature globally instead of being file-level. This loses a bit granularity for controlling such features, but given it almost never makes sense to only turn them ON/OFF individually on file level, such simplification makes it easier for debugging.

Depends on PR #177


sonalmahajan15 (approved):
Nice! Making experimental feature flags as top-level flags would definitely make it easier to experiment with these features.

PR #180, opened 2024-01-23
Fix for supporting variable declarations in rich check effects

### LINKED ISSUE #179: False positive for error return
The below code with variable declaration results in a false positive. Code rewritten to make it an assignment, such as `if v, err := retErr(); err != nil {...}` works just fine.

```
var cond bool

func retErr() (*int, error) {
      if cond {
            return nil, errors.New("some error")
      }
      return new(int), nil
}

func test() {
      var v, err = retErr()
      if err != nil {
             return
      }
      _ = *v  // FP: result 0 of `retErr()` lacking guarding; dereferenced
}
```

sonalmahajan15:
This PR fixes the handling of rich check effects by adding support for variable declarations (`ast.ValueSpec`), along with assignments (`ast.AssignStmt`).

[closes #179 ]

### DIFF
--- a/assertion/function/assertiontree/contract.go
+++ b/assertion/function/assertiontree/contract.go
@@ -22,6 +22,7 @@ import (
 
 	"go.uber.org/nilaway/annotation"
 	"go.uber.org/nilaway/util"
+	"go.uber.org/nilaway/util/asthelper"
 )
 
 // A RichCheckEffect is the fact that a certain check is associated with an effect that can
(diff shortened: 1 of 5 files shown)


assertion/function/assertiontree/contract.go
```diff
@@ -461,3 +460,17 @@ func guardExpr(rootNode *RootAssertionNode, expr TrackableExpr, guard util.Guard
...
+	case *ast.AssignStmt:
+		lhs, rhs = expr.Lhs, expr.Rhs
+	case *ast.ValueSpec:
+		for _, name := range expr.Names {
+			lhs = append(lhs, name)
+		}
+		rhs = expr.Values
+	}
+	return
+}
```
yuxincs, line 476: As we discussed internally, let's make this a function in `asthelper` since I feel this is general enough to warrant being there :)

yuxincs (approved):
Great find! Changes LGTM modulo a suggestion to move the helper function to `asthelper` util package. Approving the changes anyways 😃

assertion/function/assertiontree/contract.go
sonalmahajan15: Done :)

PR #181, opened 2024-01-24
Assignment tracking for many-to-one assignments

sonalmahajan15:
This PR adds assignment tracking for many-to-one assignments for printing informative error messages, following suit of one-to-one assignment tracking (PR #87 ).

### DIFF
--- a/assertion/function/assertiontree/backprop_util.go
+++ b/assertion/function/assertiontree/backprop_util.go
@@ -752,13 +752,15 @@ func CheckGuardOnFullTrigger(trigger annotation.FullTrigger) annotation.FullTrig
 }
 
 // addAssignmentToConsumer updates the consumer with assignment entries for informative printing of errors
-func addAssignmentToConsumer(lhs, rhs ast.Expr, pass *analysis.Pass, consumer annotation.ConsumingAnnotationTrigger) (err error) {
+func addAssignmentToConsumer(lhs, rhs ast.Expr, pass *analysis.Pass, consumer annotation.ConsumingAnnotationTrigger) error {
 	var lhsExprStr, rhsExprStr string
+	var err error
+
 	if lhsExprStr, err = asthelper.PrintExpr(lhs, pass, true /* isShortenExpr */); err != nil {
-		return err
+		return fmt.Errorf("converting LHS of assignment expr to string: %w", err)
 	}
 	if rhsExprStr, err = asthelper.PrintExpr(rhs, pass, true /* isShortenExpr */); err != nil {
-		return err
+		return fmt.Errorf("converting RHS of assignment expr to string: %w", err)
 	}
 
 	consumer.AddAssignment(annotation.Assignment{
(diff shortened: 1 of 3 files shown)


assertion/function/assertiontree/backprop_util.go
```diff
@@ -749,3 +750,22 @@ func CheckGuardOnFullTrigger(trigger annotation.FullTrigger) annotation.FullTrig
 	}
 	return trigger
 }
+
+// addAssignmentToConsumer updates the consumer with assignment entries for informative printing of errors
+func addAssignmentToConsumer(lhs, rhs ast.Expr, pass *analysis.Pass, consumer annotation.ConsumingAnnotationTrigger) (err error) {
```
yuxincs, line 755: nit: we can drop the named return since it's not necessary :)

assertion/function/assertiontree/backprop_util.go
```diff
@@ -749,3 +750,22 @@ func CheckGuardOnFullTrigger(trigger annotation.FullTrigger) annotation.FullTrig
...
+	var lhsExprStr, rhsExprStr string
+	if lhsExprStr, err = asthelper.PrintExpr(lhs, pass, true /* isShortenExpr */); err != nil {
+		return err
```
yuxincs, line 758: Can we add more context to the error here? `fmt.Errorf("adding assignment to consumer: %w", err)`

assertion/function/assertiontree/backprop_util.go
```diff
@@ -749,3 +750,22 @@ func CheckGuardOnFullTrigger(trigger annotation.FullTrigger) annotation.FullTrig
...
+	}
+	if rhsExprStr, err = asthelper.PrintExpr(rhs, pass, true /* isShortenExpr */); err != nil {
+		return err
```
yuxincs, line 761: same here :)

assertion/function/assertiontree/backprop_util.go
```diff
@@ -749,3 +750,22 @@ func CheckGuardOnFullTrigger(trigger annotation.FullTrigger) annotation.FullTrig
...
+	}
+
+	consumer.AddAssignment(annotation.Assignment{
```
yuxincs, line 764: Not something to be handled in this PR, but do you think we can actually move the logic (the `asthelper.PrintExpr`) inside the `AddAssignment` method? meaning make it take the AST nodes and do the expr shortening inside. I feel that would make the code simpler, but I may have missed some context.

yuxincs (approved):
Changes LGTM, I only have a thought for future code simplification (that may not actually be valid) for discussion.

assertion/function/assertiontree/backprop_util.go
sonalmahajan15: Good idea. Done!

assertion/function/assertiontree/backprop_util.go
sonalmahajan15: Done :)

assertion/function/assertiontree/backprop_util.go
sonalmahajan15: Agreed. Done :)

assertion/function/assertiontree/backprop_util.go
sonalmahajan15: We could. However, that would mean passing the `*analysis.Pass` to `ConsumeTrigger`, which is not required by ConsumeTrigger otherwise. Maybe we can do this refactor later if that seems to be the better approach.

PR #183, opened 2024-01-26
Minor refinement to error message format

sonalmahajan15:
This simply replaces the prefix `->` in the error message with `-`.


yuxincs (approved):
This helps render the nilability chain as lists in environments that support markdown for better visualizations. 👍

PR #185, opened 2024-01-30
Support for blank receivers in nilable receivers logic to address false negatives

### LINKED ISSUE #169: False negative with blank receivers
Below code with blank receivers results in a false negative.

```
type myErr struct{}

func (myErr) Error() string { return "myErr message" }

type S struct {
	errField error
}

func test1() {
	var e *myErr
	print(e.Error()) // FN: no error reported
}

func test2(s *S) {
        var e *myErr
        s.errField = e
	print(s.errField.Error()) // FN: no error reported
}
```

sonalmahajan15:
This PR refines the logic for nilable receivers by adding support for blank receivers, i.e., non-pointer receivers. Such receivers cannot take a nil value, and should be checked for nilability at the call site.

[Closes #169 ]

### DIFF
--- a/assertion/function/assertiontree/trusted_func.go
+++ b/assertion/function/assertiontree/trusted_func.go
@@ -145,7 +145,7 @@ var negatedSelfExpr action = func(call *ast.CallExpr, argIndex int, _ *analysis.
 	}
 }
 
-var nonnilProducer action = func(call *ast.CallExpr, argIndex int, _ *analysis.Pass) any {
+var nonnilProducer action = func(call *ast.CallExpr, _ int, _ *analysis.Pass) any {
 	return &annotation.ProduceTrigger{
 		Annotation: &annotation.TrustedFuncNonnil{ProduceTriggerNever: &annotation.ProduceTriggerNever{}},
 		Expr:       call,
--- a/assertion/function/assertiontree/util.go
+++ b/assertion/function/assertiontree/util.go
@@ -177,7 +177,7 @@ func inverseToken(t token.Token) token.Token {
 // For better performance by the caller, it also returns a boolean flag `isNoop` indicating whether
 // the returned function is a no-op
 func AddNilCheck(pass *analysis.Pass, expr ast.Expr) (trueCheck, falseCheck RootFunc, isNoop bool) {
-	noop := func(node *RootAssertionNode) {}
+	noop := func(_ *RootAssertionNode) {}
 
 	binExpr, ok := util.StripParens(expr).(*ast.BinaryExpr)
 
--- a/testdata/src/go.uber.org/receivers/inference/receivers-with-inference.go
+++ b/testdata/src/go.uber.org/receivers/inference/receivers-with-inference.go
@@ -71,7 +71,7 @@ func testInScope() {
 
 	var a *A
 	err := a.retErr()
-	print(err.Error()) // false negative, since `Error()` is nil-unsafe
+	print(err.Error()) //want "result 0 of `retErr.*`"
 }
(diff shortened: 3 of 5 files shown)


testdata/src/go.uber.org/receivers/receivers.go
```diff
@@ -95,18 +95,6 @@ func testCaller(dummy bool, i int) {
...
-	_ = s.f
-}
-
-func (*S) testNamelessPointerRecv(i int) *int {
-	return &i
-}
-
-func (S) testNamelessNonPointerRecv(i int) *int {
-	return &i
-}
```
sonalmahajan15, line 108: These methods are removed from here, since they were unused. Instead, they are updated and used in `receivers-with-inference.go` test file.

testdata/src/go.uber.org/receivers/inference/receivers-with-inference.go
```diff
@@ -111,3 +111,51 @@ func testAffiliation() {
 	// FP since affiliations are not tracked for nilable receivers
 	newI2().foo() //want "result 0 of `newI2.*`"
 }
+
```
yuxincs, line 114: Can we also add cases where the receiver is the blank identifier (`func (_ *A) foo()`)?

testdata/src/go.uber.org/receivers/receivers.go
```diff
@@ -47,8 +47,30 @@ func (s *S) nonnilRecv() {
 	_ = s.f
 }
 
-func testCaller(dummy bool, i int) {
+func (s S) nonPointerRecv() {
```
yuxincs, line 50: Same here, let's add cases where the receiver is the blank identifier :)

yuxincs (approved):
Logic LGTM! I only have a suggestion to add test cases when receivers are blank identifiers 😃

testdata/src/go.uber.org/receivers/inference/receivers-with-inference.go
sonalmahajan15: Good idea. Added.

testdata/src/go.uber.org/receivers/receivers.go
sonalmahajan15: Done :)

assertion/function/assertiontree/util.go
```diff
@@ -177,7 +177,7 @@ func inverseToken(t token.Token) token.Token {
 // For better performance by the caller, it also returns a boolean flag `isNoop` indicating whether
 // the returned function is a no-op
 func AddNilCheck(pass *analysis.Pass, expr ast.Expr) (trueCheck, falseCheck RootFunc, isNoop bool) {
-	noop := func(node *RootAssertionNode) {}
+	noop := func(_ *RootAssertionNode) {}
```
sonalmahajan15, line 180: (unrelated to this PR, but needed to add this lint fix to make the linter happy)

assertion/function/assertiontree/trusted_func.go
```diff
@@ -145,7 +145,7 @@ var negatedSelfExpr action = func(call *ast.CallExpr, argIndex int, _ *analysis.
 	}
 }
 
-var nonnilProducer action = func(call *ast.CallExpr, argIndex int, _ *analysis.Pass) any {
+var nonnilProducer action = func(call *ast.CallExpr, _ int, _ *analysis.Pass) any {
```
sonalmahajan15, line 148: (unrelated to this PR, but needed to add this lint fix to make the linter happy)

PR #186, opened 2024-01-30
Fix nondeterminism when filtering triggers for error returns

yuxincs:
We have observed nondeterministic behaviors from NilAway when analyzing the standard libraries. 

When we process the full triggers for a particular package, we first filter out the non-error triggers for the error-returning functions that are dependent on the nilability of the error, process the other triggers, and later removing some non-error triggers if the nilabilities of the error are known due to the partial processing. This helps reduce a lot of false positives.

However, we are using a set (map) to store the non-error triggers for error-returning functions since it makes later deletion faster. Eventually, we call `maps.Keys` to convert the map back to a slice to continue processing the triggers.

This brings randomness, and NilAway currently depends on stable processing orders of triggers for deterministic error reporting. 

This PR updates `FilterTriggersForErrorReturn` function to return the deleted triggers in a map instead. Then in our inference engine, we store the triggers in slices, and eventually re-iterate the ordered slice (and reference the deleted trigger map) to construct the final (ordered) filtered triggers. By doing this, we should have similar performance as the original implementation.

### DIFF
--- a/assertion/function/assertiontree/util.go
+++ b/assertion/function/assertiontree/util.go
@@ -446,12 +446,12 @@ const (
 // as well as between different modes of inference.
 //
 // FilterTriggersForErrorReturn produces two outputs:
-// (1) final set of triggers that is filtered and refined by replacing consumers
-// (2) raw set of deleted triggers
+// (1) final set of triggers that is filtered and refined by replacing consumers;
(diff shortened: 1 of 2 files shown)


sonalmahajan15 (approved):
This is nice! Glad that it fixes the non-determinism. I had not realized use of maps would create such problems when I had first implemented the logic for handling error returning functions!

PR #187, opened 2024-01-30
Add golden test and a CI job to prevent functionality regressions

yuxincs:
This PR adds a golden test for NilAway that checks the reported errors on the standard library for specified branches (default is checking the current branch vs the main branch).

Since NilAway is still at early stage of development, constantly updating / maintaining the golden test output will be a burden. Therefore, we add a separate CI job that posts the differences as a PR comment for manual reviews for now.

Implementation-wise, this PR adds a `tools` _project_ inside the NilAway project for storing [tool dependencies](https://go.dev/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) and implements the golden test logic inside `tools/cmd/golden-test`. Having a separate _project_ instead of a _package_ makes it easier to separate tool's dependencies with NilAway's dependencies for better management.

As a result, we have added this task to the main Makefile as `make golden-test`, which optionally takes arguments via the `ARGS` env var (e.g., `make golden-test ARGS="-base-branch XXX -test-branch XXX"`. We also take this opportunity to add a `make clean` task in the main Makefile.

Depends on PR #186


sonalmahajan15 (approved):
This is great! A much needed sanity checking job! 👍

PR #189, opened 2024-01-31
Add nilable receiver support for named types

sonalmahajan15:
This PR extends the nilable receiver support for methods of named types. (Earlier the support was limited to only structs.)


yuxincs (approved):
Nice fix! LGTM!

PR #191, opened 2024-02-01
Fix makefile to run test/coverage/lint on all modules in the repository

yuxincs:
By default, the `go test ./...` includes all packages and sub-packages within the current _module_. However, in PR #187 we have introduced another module (`tools`) inside this repository, which is silently ignored by test / coverage / lint tasks in our makefile. This PR fixes this by iterating over a predefined list of modules for all these tasks, making it expandable for future module additions.

Due to this fix, we have run `gofmt -s` on `tools/cmd/golden-test/main_test.go` to pass linting.


sonalmahajan15 (approved):
LGTM!

Issue #192, opened 2024-02-01
Rethink our strategies towards handling channels


yuxincs:
Currently in NilAway we forbid sending to and receiving from nilable channels by creating corresponding consumers for them. However, in practice this will not result in panics.

See the following behaviors regarding nil / closed channels (taken from [this SO post], slightly modified (https://stackoverflow.com/questions/43616434/closed-channel-vs-nil-channel)): 
> 1. A send to a nil channel blocks forever
> 2. A receive from a nil channel blocks forever
> 3. A send to a closed channel panics
> 4. A receive from a closed channel returns the zero value immediately

Only case 3 (sending to a closed channel) will result in a panic. The other cases arguably may not be within scope of NilAway's analysis. Moreover, NilAway currently only tracks nilabilities of channels, but not the states of them, making it unable to really handle case 2 and 4.

We will definitely need to rethink our strategies towards handling channels since the current approach raises a lot of false positives. For example, below is a perfectly valid and _very_ popular pattern in Go:

```go
func foo(ctx context.Context) {
  // ...
  select {
    case ctx.Done():
      return
    default:
  }
  // ...
}
```

which checks if the context is done, and if not continue to handling next rounds of business. Note that here [`ctx.Done()` return a nilable channel](https://pkg.go.dev/context#Context), but it is perfectly fine to read from it (without panics, and in this case, even without blocking forever due to the select statement).

PR #193, opened 2024-02-01
Fix NilAway FPs on channel sends/receives

### LINKED ISSUE #98: False positive for a nil channel select case
Although a direct send/receive on a nil channel can cause a panic, the following code with use of `select` is safe. However, NilAway still reports an error.

```
func foo(aChan, bChan chan string) {
	aChan <- "abc"
	bChan <- "xyz"

	if true {
		aChan = nil
	}

	select {
	case a := <-aChan:   // error: "literal `nil` uninitialized; nil channel accessed"
		print(a)
	case b := <-bChan:
		print(b)
	}
}

func test() {
	aChan := make(chan string)
	bChan := make(chan string)

	go foo(aChan, bChan)
}
```

### LINKED ISSUE #167: Receiving from channel is mistakenly checked
as the spec says "Receiving from a nil channel blocks forever." https://go.dev/ref/spec#Receive_operator

But in code we treat it as panic

https://github.com/uber-go/nilaway/blob/2f6a74d7c0e26d87215dfbda2c9ab5acbb64b11c/assertion/function/assertiontree/root_assertion_node.go#L812C1-L813C1

This solves https://github.com/uber-go/nilaway/issues/98

yuxincs:
See issue #192 for a more detailed discussion on handling channels in NilAway. To make this PR more self-contained, below is a partial copy of the discussion there:

See the following behaviors regarding nil / closed channels (taken from [this SO post](https://stackoverflow.com/questions/43616434/closed-channel-vs-nil-channel), slightly modified): 
> 1. A send to a nil channel blocks forever
> 2. A receive from a nil channel blocks forever
> 3. A send to a closed channel panics
> 4. A receive from a closed channel returns the zero value immediately

Only case 3 (sending to a closed channel) will result in a panic. The other cases arguably may not be within scope of NilAway's analysis. Moreover, NilAway currently only tracks nilabilities of channels, but not the states of them, making it unable to really handle case 2 and 4.

Since this introduces a lot more FPs than TPs (if any), this PR removes logic to require sending to/receiving from nonnil channels. We leave properly handling channels as future work (especially around tracking states of channels).

The test cases have been updated to reflect this. Specifically, we kept the test cases there to now serve as negative cases in case for future development of such a support.

Fixes #98 
Fixes #167

### DIFF
--- a/assertion/function/assertiontree/backprop.go
+++ b/assertion/function/assertiontree/backprop.go
@@ -116,13 +116,13 @@ func backpropAcrossNode(rootNode *RootAssertionNode, node ast.Node) error {
 // backpropAcrossSend handles backpropagation for send statements. It is designed to be called from
 // backpropAcrossNode as a special handler.
 func backpropAcrossSend(rootNode *RootAssertionNode, node *ast.SendStmt) error {
-	// Added this consumer since sending over a nil channel can cause panic
-	rootNode.AddConsumption(&annotation.ConsumeTrigger{
-		Annotation: &annotation.ChanAccess{ConsumeTriggerTautology: &annotation.ConsumeTriggerTautology{}},
-		Expr:       node.Chan,
-		Guards:     util.NoGuards(),
-	})
-
+	// Note that for channel sends, we have:
+	// (1) A send to a nil channel blocks forever;
+	// (2) A send to a closed channel panics.
+	// (1) falls out of scope for NilAway and hence we do not create a consumer here for the
+	// channel variable. For (2), since we do not track the state of the channels, we currently
+	// can not support it.
+	// TODO: rethink our strategy of handling channels (#192).
 	consumer, err := exprAsAssignmentConsumer(rootNode, node, nil)
 	if err != nil {
 		return err
--- a/assertion/function/assertiontree/backprop.go
+++ b/assertion/function/assertiontree/backprop.go
@@ -121,7 +121,7 @@ func backpropAcrossSend(rootNode *RootAssertionNode, node *ast.SendStmt) error {
 	// (2) A send to a closed channel panics.
 	// (1) falls out of scope for NilAway and hence we do not create a consumer here for the
 	// channel variable. For (2), since we do not track the state of the channels, we currently
-	// can not support it.
+	// cannot support it.
 	// TODO: rethink our strategy of handling channels (#192).
 	consumer, err := exprAsAssignmentConsumer(rootNode, node, nil)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


assertion/function/assertiontree/backprop.go
```diff
@@ -116,13 +116,13 @@ func backpropAcrossNode(rootNode *RootAssertionNode, node ast.Node) error {
...
-		Expr:       node.Chan,
-		Guards:     util.NoGuards(),
-	})
-
+	// Note that for channel sends, we have:
+	// (1) A send to a nil channel blocks forever;
+	// (2) A send to a closed channel panics.
+	// (1) falls out of scope for NilAway and hence we do not create a consumer here for the
+	// channel variable. For (2), since we do not track the state of the channels, we currently
+	// can not support it.
```
sonalmahajan15, line 124:
```suggestion
	// cannot support it.
```

assertion/function/assertiontree/root_assertion_node.go
```diff
@@ -807,15 +807,14 @@ func (r *RootAssertionNode) AddComputation(expr ast.Expr) {
...
-				Guards:     util.NoGuards(),
-			})
-		}
+		// Note if expr.Op == token.ARROW it represents a channel receive (<-X), and we have:
+		// (1) A receive from a nil channel blocks forever;
+		// (2) A receive from a closed channel returns the zero value immediately.
+		// (1) falls out of scope of NilAway, and we have a lot of valid Go code that receives
+		// from nil channels (e.g., select statements with nilable channels). So we do not create
+		// consumer for the channel variable here. For (2), since we currently do not track the
+		// state of channels, we currently can not support it either.
```
sonalmahajan15, line 816:
```suggestion
		// state of channels, we currently cannot support it either.
```

testdata/src/go.uber.org/errormessage/errormessage.go
```diff
@@ -101,28 +101,22 @@ func test9(m map[int]*int) {
...
+// nilable(nilableChan) nonnil(nonnilDeeplyNonnilChan, <-nonnilDeeplyNonnilChan)
+func test10(nilableChan chan *int, nonnilDeeplyNonnilChan chan *int) {
+	x := 1
+	nilableChan <- &x
+	// Sending nilable values to nonnil and deeply nonnil channels is not OK.
+	var y *int
+	nonnilDeeplyNonnilChan <- y //want "`y` assigned deeply into parameter arg `nonnilDeeplyNonnilChan`"
 }
 
+// nilable(s)
```
sonalmahajan15, line 113: Unrelated to the PR, I guess, but a welcome change :)

sonalmahajan15 (approved):
As discussed internally, we did agree upon this strategy to address false positives. LGTM!

testdata/src/go.uber.org/errormessage/errormessage.go
yuxincs: Yes :)

PR #194, opened 2024-02-01
Fix golden test to handle "HEAD" being the branch name

yuxincs:
Previously in our golden test we were not specifically handling branch names being "HEAD". For example, if the test branch is "HEAD" and we first check out to the base branch and later the test branch ("HEAD") for golden testing, we would be testing against the base branch itself, giving incorrect results.

This PR fixes that by checking out the short commit SHA instead. The logic for switching back to the original branch is also updated to handle "HEAD" (if we were in a detached state before the golden test).


sonalmahajan15 (approved):
Good catch!

PR #195, opened 2024-02-08
Add integration test framework with different linter drivers

yuxincs:
We have a extensive test suite in NilAway using the `analysistest` test framework, where we use `//want` strings to test if an expected diagnostic is properly emitted by NilAway. However, it has two major limitations that make it undesirable to ensure real-world behaviors:

* It does not read the "want" comments in the upstream package when analyzing downstream packages. 
  This makes it impossible to test our multi-package inference algorithm: NilAway might report an 
  error for an upstream package while analyzing downstream packages, but we cannot write expected 
  error strings for the framework to verify;

* For bazel/nogo, the package loader and [`Fact`](https://pkg.go.dev/golang.org/x/tools/go/analysis#hdr-Modular_analysis_with_Facts) caching behavior is slightly different from 
  bazel (which under the hood uses a custom nogo runner that is different from the `analysistest` 
  runner), and there is no guarantee that they will not diverge further in the future.

This PR adds a framework for testing cross-package inference, and with different linter drivers (with the same `//want` string to specify expected diagnostics). As an initial step, only standalone checker is supported, support for bazel/nogo will be added in a follow-up PR. 

A CI job is also added. We also take this opportunity to add `goleak` to the tests within the `tools` module.


tools/cmd/integration-test/main_test.go
```diff
@@ -0,0 +1,93 @@
+package main
```
sonalmahajan15, line 1: Appreciate the comprehensive tests! :)

tools/cmd/integration-test/standalone.go
```diff
@@ -0,0 +1,66 @@
+package main
```
sonalmahajan15, line 1: Same here. Appreciate the tests! :)

sonalmahajan15 (approved):
Integration testing was much needed to make sure new features work seamlessly. Thanks for implementing this!

PR #196, opened 2024-02-08
Upgrade checkout actions version

yuxincs:
This PR upgrades the `actions/checkout` and `actions/cache` versions to avoid the deprecation warnings in CI runs.

We should also consider configuring dependabot to do this in the future, but for now let's merge the upgrades.


sonalmahajan15 (approved):
LGTM

PR #88, opened 2023-10-20
Track interprocedural deep nilability

sonalmahajan15:
This PR addresses false negatives related to interprocedural deep nilability by adding support for the same.

[closes #61]
[depends on #60]

PR #197, opened 2024-02-09
Add missing gob encodings

sonalmahajan15:
This PR the registers gob encodings for the newly created consumer trigger structs in PR #88 .

### DIFF
--- a/inference/engine.go
+++ b/inference/engine.go
@@ -525,6 +525,8 @@ func GobRegister() {
 	gob.RegisterName(nextStr(), annotation.GlobalVarAssignDeepPrestring{})
 	gob.RegisterName(nextStr(), annotation.LocalVarAssignDeepPrestring{})
 	gob.RegisterName(nextStr(), annotation.ChanSendPrestring{})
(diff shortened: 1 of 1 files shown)


omar-scio:
I just ran into this last night and put up https://github.com/uber-go/nilaway/pull/198 before I saw yours already existed.

Might I add that there be some sort of unittest to catch these? Otherwise, it's fairly easy for a developer to make that mistake right?

sonalmahajan15:
> I just ran into this last night and put up #198 before I saw yours already existed.
> 
> Might I add that there be some sort of unittest to catch these? Otherwise, it's fairly easy for a developer to make that mistake right?

Oops, sorry about PR #198 !

Absolutely agreed, I had the same thought that we should add unit tests to catch such mishaps. In fact, I was planning to work on it next week (issue #200). But if you are still interested in making a first contribution to NilAway, I'd be happy to help you with that! :)

yuxincs (approved):
Important fix! Let's try to find a way to do unit tests for this to prevent future regressions!

PR #201, opened 2024-02-13
Upgrade Go version to 1.21

yuxincs:
Since [Go 1.22 was released on 2/6](https://tip.golang.org/doc/go1.22), we are now upgrading the minimum Go version requirement to 1.21 given our version support policy (i.e., the last two major Go versions, which is the same as the official Go version support).

This would unlock us to use a few new features introduced in Go 1.21 for better performance / maintainability (which will be done in follow-up PRs).

We also take this opportunity to upgrade our dependencies.


sonalmahajan15 (approved):
LGTM :)

PR #202, opened 2024-02-13
Optimize codebase with features from Go 1.21

yuxincs:
This PR optimizes the codebase using the new features introduced in Go 1.21 since we have bumped our minimum required version.

Specifically, we are:

(1) leveraging the builtin `clear` to clear the entire slice instead of allocating a new one in each iteration of backpropagation.
(2) Replacing the helper functions in `x/exp` for slices and maps with the corresponding one from standard library. By doing this, we have removed the dependency for `x/exp`.
(3) Renaming a local variable `max` to avoid collisions with the new builtin function `max`.
(4) Replacing `strings.Compare` with the generic version `cmp.Compare` from standard libraries.
(5) Using `token.File.Lines()` for better performance when checking if a file is "fake" (i.e., imported from archive).

All these changes bring slight performance gains and/or better code readability.

Depends on PR #201

### DIFF
--- a/annotation/helper_test.go
+++ b/annotation/helper_test.go
@@ -24,7 +24,6 @@ import (
 
 	"github.com/stretchr/testify/mock"
 	"go.uber.org/nilaway/util"
-	"golang.org/x/exp/maps"
 	"golang.org/x/tools/go/packages"
 )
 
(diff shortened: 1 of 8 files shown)


sonalmahajan15 (approved):
All welcome changes! :)

PR #203, opened 2024-02-13
Add unit tests for gob encoding of `inference.InferredMap`

### LINKED ISSUE #200: Write unit tests for ensuring Gob registration
Write unit tests to catch if any of the consumer or producer triggers are missing Gob registration to avoid the case that happened with PR #197 .

yuxincs:
In PR #88, we accidentally missed registering a type to gob encoding for the `inference.InferredMap` and it wasn't actually caught by the unit tests. This is because the `analysistest` framework does not check this by default. 

We have implement such checking by encoding and immediately decoding the `inference.InferredMap` whenever the analysis is finished under testing environment to catch these kinds of mistakes early in the development process.

We also take this opportunity to refactor the top-level nilaway tests to be [table-driven tests](https://github.com/uber-go/guide/blob/master/style.md#test-tables) to avoid code duplication.

Fixes #200 

Depends on PR #202 and PR #197

### DIFF
--- a/inference/inferred_map.go
+++ b/inference/inferred_map.go
@@ -19,6 +19,7 @@ import (
 	"encoding/gob"
 	"errors"
 	"go/types"
+	"testing"
 
 	"github.com/klauspost/compress/s2"
 	"go.uber.org/nilaway/annotation"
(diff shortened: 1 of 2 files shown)


inference/inferred_map.go
```diff
@@ -109,6 +110,28 @@ func (i *InferredMap) Export(pass *analysis.Pass) {
...
+	if testing.Testing() {
+		var buf bytes.Buffer
+		if err := gob.NewEncoder(&buf).Encode(i); err != nil {
+			panic(err)
+		}
+		var m *InferredMap
+		if err := gob.NewDecoder(&buf).Decode(&m); err != nil {
+			panic(err)
+		}
+	}
```
sonalmahajan15, line 133: Makes sense to me. Unfortunately, an awkward place to insert testing logic, but I totally get why we had to go this route. Also, the detailed explanation in the comment  helps.

nilaway_test.go
```diff
@@ -25,205 +25,57 @@ import (
...
-	analysistest.Run(t, testdata, Analyzer, "go.uber.org/functioncontracts", "go.uber.org/functioncontracts/inference")
-}
-
-func TestConstants(t *testing.T) {
-	t.Parallel()
-
-	testdata := analysistest.TestData()
-	analysistest.Run(t, testdata, Analyzer, "go.uber.org/consts")
-}
+func TestNilAway(t *testing.T) {
```
sonalmahajan15, line 28: I like this change from the point of view of test management 👍

nilaway_test.go
```diff
@@ -25,205 +25,57 @@ import (
...
+		{name: "Generics", patterns: []string{"go.uber.org/generics"}},
+		{name: "FunctionContracts", patterns: []string{"go.uber.org/functioncontracts", "go.uber.org/functioncontracts/inference"}},
+		{name: "Constants", patterns: []string{"go.uber.org/consts"}},
+		{name: "ErrorMessage", patterns: []string{"go.uber.org/errormessage"}},
+	}
 
-func TestErrorMessage(t *testing.T) {
-	t.Parallel()
+	for _, tt := range tests {
+		tt := tt
```
sonalmahajan15, line 71: Is this shadowing really needed?

sonalmahajan15 (approved):
LGTM :)

nilaway_test.go
yuxincs: Yes, it's a common pattern for Go, since loop variable is per loop not per iteration, so we must shadow the variable such that the parallel test is accessing the correct `tt`.

Luckily, Go finally got rid of this weirdness in Go 1.22: https://tip.golang.org/doc/go1.22 so we don't have to use this trick once we require Go 1.22 (i.e., in another 6 months) :)

PR #204, opened 2024-02-16
Fix for infinite loops

sonalmahajan15:
This PR adds support for correctly analyzing `while` loops, particularly infinite loops, that were causing false negatives.

### DIFF
--- a/assertion/function/assertiontree/backprop_util.go
+++ b/assertion/function/assertiontree/backprop_util.go
@@ -664,9 +664,11 @@ func blocksAndPreprocessingFromCFG(
 		Index: int32(numBlocks),
 		Live:  true,
 	})
-	// link all returning blocks to the "return" block
+	// add "return" block as a successor for:
+	// - all returning blocks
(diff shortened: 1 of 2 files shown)


yuxincs (approved):
https://github.com/golang/go/blob/b634f6fdcbebee23b7da709a243f3db217b64776/src/runtime/proc.go#L296-L299

I'm scratching my head off trying to understand what the false negative from the stdlib is doing (seems unreachable, but not sure why Go folks put it there in the first place).....

Anyways, this is definitely a very rare case, but we should add support nonetheless. Thanks for the fix!

sonalmahajan15:
> https://github.com/golang/go/blob/b634f6fdcbebee23b7da709a243f3db217b64776/src/runtime/proc.go#L296-L299
> 
> I'm scratching my head off trying to understand what the false negative from the stdlib is doing (seems unreachable, but not sure why Go folks put it there in the first place).....
I know! Even did spend a long time trying to reason about it. 🤷‍♀️

PR #205, opened 2024-02-16
Add a helper function that wraps the `Analyzer.Run` for panics

yuxincs:
We strive to _never_ panic in NilAway: we have deferred function calls that recover from any panics and converts it to an error, which eventually gets converted to a diagnostic with "INTERNAL PANIC" prefix for users to be aware (but not to stop the build/linting process etc.).

However, we were adding the same deferred function block for _every_ sub-analyzer we have, which introduces maintainability burden.

This PR simplifies such handling by introducing a helper function that wraps around the `Analyzer.Run` function, which:

* handles panics;
* puts the error in a field in `Result` struct to be sent via the return value, such that any error won't stop the analysis process. The upper-level analyzer will decide what to do with the error (for future graceful recovery). For now, the upper-level analyzers simply early return if there is an error, and the accumulation analyzer will convert the errors to be diagnostics.

Tests that intentionally cause a panic inside the analyzers are also added for each sub-analyzer to ensure that it won't crash the program (i.e., it is properly recovered and converted to errors).

We also take this opportunity to add a few missing license headers to some files.

### DIFF
--- a/accumulation/analyzer.go
+++ b/accumulation/analyzer.go
@@ -18,6 +18,7 @@
 package accumulation
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 	"runtime/debug"
@@ -28,6 +29,7 @@ import (
 	"go.uber.org/nilaway/config"
 	"go.uber.org/nilaway/diagnostic"
 	"go.uber.org/nilaway/inference"
+	"go.uber.org/nilaway/util/analysishelper"
 	"golang.org/x/tools/go/analysis"
 )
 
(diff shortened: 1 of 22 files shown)


accumulation/analyzer.go
```diff
@@ -164,16 +160,6 @@ func run(pass *analysis.Pass) (result interface{}, _ error) {
 	return diagnostics, nil
 }
 
-// errorsToDiagnostics converts the internal errors to a slice of analysis.Diagnostic to be reported.
-func errorsToDiagnostics(errs []error) []analysis.Diagnostic {
```
yuxincs, line 168: inlined due to the simplification of error handling logic

sonalmahajan15 (approved):
The changes make sense to me. They help improve the functionality as well as the readability of code. Also, appreciate the tests for different analyzers. :)

PR #209, opened 2024-02-23
Upgrade codecov action to v4

yuxincs:
This PR upgrades our codecov action to v4. A side effect of the upgrade is that tokenless uploads are no longer supported by the action. Therefore, we also fixes the config and give it a token stored in repository secrets.


sonalmahajan15 (approved):
Thanks for fixing this! :)

PR #210, opened 2024-02-23
Fix panics from contract analyzer

yuxincs:
This PR fixes a panic from the contract analyzer where it is erroneously panicking if there are no blocks in the `*ssa.Function`. For external functions like builtin function or other C functions implemented outside Go, since we do not actually have Go sources for them, their SSA form will not have any blocks. 

This PR removes the panic and simply ignore those functions, and test cases for such functions (where functions are declared without bodies) are added to prevent future regressions.


sonalmahajan15 (approved):
Good catch!

PR #216, opened 2024-03-12
Sort conflicts such that diagnostics are always grouped under the first diagnostic

yuxincs:
We had user reports internally that the diagnostics are not always grouped under the first diagnostic, making it a little bit annoying to see a diagnostic at L35 and reports "with similar errors at L5". 

This PR sorts the conflicts in the diagnostic engine before returning the converted diagnostics, such that diagnostics are always grouped under the first diagnostic.

### DIFF
--- a/diagnostic/conflict.go
+++ b/diagnostic/conflict.go
@@ -21,9 +21,12 @@ import (
 )
 
 type conflict struct {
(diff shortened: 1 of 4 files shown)


sonalmahajan15 (approved):
I like the improved UX. Nice work! :)

PR #218, opened 2024-03-18
Fix for chained field accesses with negative nil checks in non-conditional flows

### LINKED ISSUE #217: False positive for negative nil checks with chained field accesses
Below test case with chained field access results false positives.

```
// nilable(f)
type X struct {
   f *G
}

// nilable(g)
type G struct {
   g *H
}

type H struct {
   h int
}

// nilable(x)
func foo(x *X) bool {
   return x != nil && x.f != nil && x.f.g != nil && x.f.g.h == 4 // FPs reported here
} 
```

sonalmahajan15:
This PR enhances the support for tracking negative nil checks in non-conditional flows, particularly in the case of chained field accesses. E.g., `return x != nil && x.f != nil && x.f.g != nil && x.f.g.h == 4` reported false positives for the field accesses `x.f.g` and `x.f.g.h`. This PR fixes that problem, thereby reducing false positives.

[closes #217 ]

### DIFF
--- a/assertion/function/assertiontree/root_assertion_node.go
+++ b/assertion/function/assertiontree/root_assertion_node.go
@@ -553,12 +553,19 @@ func (r *RootAssertionNode) AddComputation(expr ast.Expr) {
 		// Process the binary expression `X op Y` in reverse, i.e., add consumers for Y first and then X
 		r.AddComputation(expr.Y)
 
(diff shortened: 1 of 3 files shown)


yuxincs (approved):
Clever solution on recursively processing `expr.Y`! Let's add a few more comments there for future readers, but the logic LGTM :)

PR #219, opened 2024-03-19
Truncate output of golden test output due to GH limit

yuxincs:
GitHub has a hard limit on the length of the comment to be 65535. Since our golden test results are posted as a GH comment in each PR, sometimes if the change is too large we hit such limit.

This PR replaces the middle part of the diff with `... (truncated) ...` in the comment if it gets too large.


sonalmahajan15 (approved):
Interesting. Glad you found and addressed this problem! :)

PR #222, opened 2024-03-21
Fix breaking changes from golangci-lint github actions

yuxincs:
This PR fixes the breaking changes from golangci-lint github actions.

Moreover, instead of always using the latest version of golangci-lint, we now bundle a `.golangci.version` file that specifies the version we use for `golangci-lint` to avoid future breaks.

With the introduction of this version lock file, we also refactored our Makefile to automatically download the desired version of `golangci-lint` and store in inside the `bin` folder of the nilaway repository. This ensures that even local linting happens with our desired golangci-lint version (and will automatically redownloads the binary if the version changes in the future).

We also take this opportunity to upgrade `actions/setup-go` in our CI jobs


sonalmahajan15 (approved):
LGTM! :)

PR #223, opened 2024-03-21
Fix for loop range handling of basic types and test refactoring

### LINKED ISSUE #221: False positive for range over integers
From Go 1.22, range over intergers is supported (https://tip.golang.org/doc/go1.22), but `nilaway` returns the error 
```
backpropagation across node (...) of type *ast.AssignStmt failed for reason: unrecognized type of rhs in range statement: int
```

sonalmahajan15:
[Go 1.22](https://tip.golang.org/doc/go1.22) now allows ranging over integers. NilAway was not handling this correctly and throwing a panic. This PR fixes that problem. I also took the opportunity to create a new `looprange` test file that now hosts all the range related test cases pulled in from other files and new test cases for checking the support for basic types.

[closes #221 ]


yuxincs (approved):
Nice fix and test refactors!

yuxincs:
Oh, I think you need to separate the tests for the new syntax and only run the tests for >1.22 until our minimum supported version is 1.22.

Issue #92, opened 2023-11-02
Fix false positive for a corner case of short-circuiting expressions


yuxincs:
NilAway is complaining about the latter dereference in the short-circuiting expressions:

```
func foo() bool {
  return s == nil || *s == ""
}
```

We have support for short-circuiting expressions in NilAway, but this is a corner case that we somehow missed. We should add support for it.

daboross:
Same here, for

```
type AType struct {
	Field int
}

func F1(a *AType, b *AType) bool {
	return a != nil && b != nil && a.Field != b.Field
}

func TestF1(t *testing.T) {
	if F1(nil, nil) {
		t.Fatalf("f1 should not accept nil, nil")
	}
}```

**Comment by mavirn** [comment]
Similar experience here for this code block where short circuiting will prevent accessing a field of a nil item. 

```go
func (r *Room) getInfoAboutRoom() {
    hasField := r.Obj != nil && r.Obj.Field != nil
    ...
```

davecgh:
It also is giving false positives for short circuiting slice indexing behind a length check which is very common and convenient operation because it returns 0 for both `nil` slices as well as non-nil empty slices meaning only a single check is necessary to cover both cases.

Here is a simple program to reproduce the false positive:

```Go
package main

import "fmt"

func isFirstByteZero(s []byte) bool {
        return len(s) > 0 && s[0] == 0
}

func main() {
        fmt.Println(isFirstByteZero(nil))
}
```

```
$ nilaway ./...
nilawayfp\main.go:6:23: error: Potential nil panic detected. Observed nil flow from source to dereference point:
        -> nilawayfp\main.go:10:30: literal `nil` passed as arg `s` to `isFirstByteZero()`
        -> nilawayfp\main.go:6:23: function parameter `s` sliced into

PR #225, opened 2024-03-25
Revisiting non-conditional short-circuit handling

sonalmahajan15:
This PR revisits the non-conditional short-circuit handling. Following are the changes:

- The main change is to repurpose the `AddNilCheck` function to trigger the handling, instead of the earlier custom (and duplicated) implementation in `asNilCheckExpr`. 
- `AddNilCheck` is strengthened to also be able to handle negations. E.g., `!(x == nil)`
- The added benefit of using `AddNilCheck()` is also that it already has support for `len()` checks, which the non-conditional handling can leverage right away. A new extensive set of test cases for len checks (as pointed in issue #92) is also added.


assertion/function/assertiontree/util.go
```diff
@@ -179,10 +179,29 @@ func inverseToken(t token.Token) token.Token {
...
+		// and false branches produced by the binary expression. For example, if`!(v != nil)`, then AddNilCheck on the inner
+		// expression `(v != nil)` returns trueCheck: produceNegativeNilCheck, falseCheck: noop, isNoop = false, implying a
+		// negative nil check for the true branch. But since it is preceded with a negation (!), the below code
+		// interchanges the true and false branches, and returns trueCheck: noop, falseCheck: produceNegativeNilCheck,
+		// implying a negative nil check for the false branch.
+		if e.Op == token.NOT {
+			trueNilCheck, falseNilCheck, isNoop := AddNilCheck(pass, e.X)
+			return falseNilCheck, trueNilCheck, isNoop
+		}
+	}
```
yuxincs, line 200: nit: seems simpler to get rid of the switch case and just treat the `!(expr)` as a special case and early return, right?

```go
if e, ok := expr.(*ast.UnaryExpr); ok && e.Op == token.NOT {
  return ...
}

binExpr, ok := expr
if !ok {
  return ...
}
```

testdata/src/go.uber.org/nilcheck/nonconditionalflow.go
```diff
@@ -202,3 +202,50 @@ func testChainedAccesses(x *X, i int) bool {
 	}
 	return false
 }
+
+// ---- test len checks ----
```
yuxincs, line 206: Nice addition to the test suite!!

yuxincs (approved):
Great find and nice tech debt cleanup!

I only have a nit for readability (maybe), feel free to disagree.

assertion/function/assertiontree/util.go
sonalmahajan15: Fair enough. Updated the code to replace the `switch`.

PR #230, opened 2024-04-09
Move grouping flag to the top

sonalmahajan15:
This PR creates and moves the flag for grouping similar error messages to the top.


yuxincs (approved):
It's great that we expose this option to the end-user. Could we also update our wiki for this? https://github.com/uber-go/nilaway/wiki/Configuration

PR #229, opened 2024-04-08
Fix grouping of error messages when authentic producer expr is unavailable

sonalmahajan15:
This PR fixes the incorrect grouping of error messages, which specifically occurs in the case of `Single Assertion Conflicts` when the authentic producer expression is unavailable.

PR #232, opened 2024-04-16
Refine grouping condition for single assertion conflict

sonalmahajan15:
This PR further refines the grouping condition updated in PR #229 to add function name to the key. This is particularly useful when the text of the producer and consumer messages is not sufficient to distinguish the conflicts.


diagnostic/conflict.go
```diff
@@ -72,6 +77,36 @@ func groupConflicts(allConflicts []conflict) []conflict {
 			key = p.producerRepr + ";" + p.consumerRepr
 			if p.producerPosition.IsValid() {
 				key = p.producerPosition.String() + ": " + p.producerRepr
+			} else {
```
yuxincs, line 80: Can we add some comments here listing an example where we would hit this case for future readers? :)

diagnostic/conflict.go
```diff
@@ -72,6 +77,36 @@ func groupConflicts(allConflicts []conflict) []conflict {
...
+							functionStart := pass.Fset.Position(fd.Pos()).Offset
+							functionEnd := pass.Fset.Position(fd.End()).Offset
+							if c.position.Offset >= functionStart && c.position.Offset <= functionEnd {
+								key = fd.Name.Name + ":" + key
+								return false
+							}
+						}
+						return true
+					})
+				}
```
yuxincs, line 109: I think this AST search can be simplified to just a single pass over `ast.File` node since function declarations are top-level decls and can only appear under `ast.File` node: https://go.dev/ref/spec#Declarations_and_scope

It would save some further traversals :)

yuxincs (approved):
Overall looks good to me (we can also refine this logic to perhaps include a full AST path from root to the diagnostic to further distinguish the diagnostics, but the function name heuristics work for majority of the cases).

I only have one request for some comments and an small optimization :)

diagnostic/conflict.go
sonalmahajan15: Done :)

diagnostic/conflict.go
sonalmahajan15: Great idea! Updated the code accordingly.

sonalmahajan15:
> Overall looks good to me (we can also refine this logic to perhaps include a full AST path from root to the diagnostic to further distinguish the diagnostics, but the function name heuristics work for majority of the cases).
> 
> I only have one request for some comments and an small optimization :)

Agreed, we can decide to refine the logic further to include the full path if the need arises in the future.

Addressed your comments :)

PR #233, opened 2024-04-19
Support for non-literals in map accesses

### LINKED ISSUE #84: Nilaway gives false-positive on nested map initialization
```go
regroupedData := make(map[string]map[string]map[string][]string)
for _, record := range recordSet {
	if regroupedData[record.lvl1] == nil {
		regroupedData[record.lvl1] = make(map[string]map[string][]string)
	}
	for lvl3, data := range record.Data {
		if regroupedData[record.lvl1][data.lvl2] == nil {
			// Nilaway gives false positive on this line, even though every previous nest level is initialized
			regroupedData[record.lvl1][data.lvl2] = make(map[string][]string)
		}
		regroupedData[record.lvl1][data.lvl2][lvl3] = data.CompromisedData
	}
}
```

				
error: Potential nil panic detected. Observed nil flow from source to dereference point: -> code/code.go: deep read from variable `regroupedData` lacking guarding; written to at an index

### LINKED ISSUE #111: False positive in accessing nested maps
NilAway reports a false positive when accessing nested maps.

```
func test(m map[string]map[string]*int, k1 string, k2 string) {
	if _, ok := m[k1]; !ok {
		m[k1] = map[string]*int{}
	}
	m[k1][k2] = new(int) // error: "deep read from parameter `m` lacking guarding; written to at an index"
}
```

The above code rewritten to the following works fine, though.
```
func test(m map[string]map[string]*int, k1 string, k2 string) {
	if _, ok := m[k1]; !ok {
		m[k1] = map[string]*int{}
	}

        if v, ok := m[k1]; ok {
	        v[k2] = new(int)
        }
}
```

### LINKED ISSUE #206: False positive for self referencing type in for loop
```go
type Node struct {
	children map[rune]*Node
}

func Insert(root *Node, key string) {
	if root == nil {
		return
	}
	current := root
	for _, v := range key {
		if current.children == nil {
			current.children = make(map[rune]*Node)
		}
		if current.children[v] == nil {
			current.children[v] = &Node{
				children: make(map[rune]*Node),
			}
		}
		
		// deep read from field `children` lacking guarding; accessed field `children` via the assignment
		current = current.children[v] 
	}
}
```

sonalmahajan15:
This PR adds support for effectively tracking non-literal map accesses (e.g., `mp[i]`). Earlier NilAway could support only the `ok-form` style or literal map accesses, such as shown in the examples below, and report false positives for non-literal map accesses.

Examples:
1. ok-form style
```
if v, ok := mp[i]; ok {
      return *v
}
```
2. Literal map access
```
if mp[0] != nil {
     return *mp[0]
}
```
3. Non-literal map access
```
if mp[i] != nil {
      return *mp[i]
}
```

This PR adds support for examples like 3., which are commonly found in real-world code. This support is added by extending stable expressions to handle index expressions of the following common forms:
- Index is a variable (e.g., `m[i]`)
- Index is a built-in function (e.g., `m[len(m)-1]`)
- Index is a field selector chain (e.g., `m[g.h.i]`)

[closes #111 ]
[closes #84 ]
[closes #206 ]

### DIFF
--- a/assertion/function/assertiontree/parse_expr_producer.go
+++ b/assertion/function/assertiontree/parse_expr_producer.go
@@ -326,16 +326,42 @@ func (r *RootAssertionNode) ParseExprAsProducer(expr ast.Expr, doNotTrack bool) 
 			return nil, parseDeepRead(recv, expr.X, expr, rproducers)
 		}
 		if recv != nil {
-			// receiver is trackable
-			if r.isStable(expr.Index) {
-				// receiver is trackable and index is stable, so return an augmented path
+			// X part of the expression is trackable. Now we need to check if the index is stable or trackable.
+			// If the index is not stable, it is still considered trackable if it falls into any of these categories:
+			// - Index is a variable (e.g., `m[i]`)
+			// - Index is a built-in function (e.g., `m[len(m)-1]`)
+			// - Index is a field selector chain (e.g., `m[g.h.i]`)
+			var isIndexTrackable func(expr ast.Expr) bool
+			isIndexTrackable = func(expr ast.Expr) bool {
+				switch index := expr.(type) {
+				case *ast.Ident:
+					return r.isVariable(index) || r.isStable(expr)
+				case *ast.BinaryExpr:
+					if index.Op == token.SUB || index.Op == token.ADD {
+						return isIndexTrackable(index.X) && isIndexTrackable(index.Y)
+					}
+					return false
+				case *ast.CallExpr:
+					if fun, ok := index.Fun.(*ast.Ident); ok {
+						return r.isBuiltIn(fun)
+					}
+					return false
+				case *ast.SelectorExpr:
+					return util.IsFieldSelectorChain(index)
+				default:
+					return r.isStable(expr)
+				}
+			}
+
+			if isIndexTrackable(expr.Index) {
+				// receiver is trackable and index is stable or trackable, so return an augmented path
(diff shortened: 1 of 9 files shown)


assertion/function/assertiontree/parse_expr_producer.go
```diff
@@ -326,16 +326,42 @@ func (r *RootAssertionNode) ParseExprAsProducer(expr ast.Expr, doNotTrack bool)
...
 		}
 		if recv != nil {
-			// receiver is trackable
-			if r.isStable(expr.Index) {
-				// receiver is trackable and index is stable, so return an augmented path
+			// X part of the expression is trackable. Now we need to check if the index is stable or trackable.
+			// If the index is not stable, it is still considered trackable if it falls into any of these categories:
+			// - Index is a variable (e.g., `m[i]`)
+			// - Index is a built-in function (e.g., `m[len(m)-1]`)
+			// - Index is a field selector chain (e.g., `m[g.h.i]`)
```
yuxincs, line 333: nit: Can we add a TODO here indicating that we plan to add an analyzer and only consider them trackable if there are no re-assignments?

It's just not very obvious why we should consider them trackable here ;)

yuxincs (approved):
Overall LGTM per our discussion, let's add some TODOs and more FN cases for the future analyzer :)

yuxincs (approved):
Thanks! my comments are addressed :)

PR #234, opened 2024-04-23
Implement the plugin interface for golangci-lint

### LINKED ISSUE #175: Implement our own linter driver for NilAway as a standalone checker
There are 3 ways (although one of them is broken) to run NilAway:

(1) standalone checker: we use the standard linter driver from x/tools for running NilAway, such that NilAway can be run as a standalone executable.
(2) bazel/nogo: NilAway is configured as nogo analyzer and executed by nogo.
(3) golangci-lint: NilAway is configured as a linter in golangci-lint and executed by it.

(1) runs NilAway completely in memory, where NilAway has to re-analyze unmodified packages in consecutive runs, plus consuming a huge amount of memory sinc all facts are always stored instead of being fetched on demand (see #127, #161 for resource consumption when running NilAway). (2) and (3) offer fact caching, making them much better options for running NilAway given the sophistication of the analyses that NilAway does. 

However, (2) requires users to use bazel to build their Go projects, and [(3) is broken and likely won't work for some time ](https://github.com/uber-go/nilaway/tree/4335a432c225cf9de46a3baf2db3935bc115a376?tab=readme-ov-file#golangci-lint)(until NilAway is mature enough). Both of these "better" options seem to be out of reach for the majority of users. Therefore, most are simply using (1) and living with the performance penalties.

Given the demand, we might try to implement our own minimum caching linter driver that handles scheduling of NilAway sub-analyzers, as well as storing fact files to the filesystem (as well as loading them on-demand). We may need to re-invent some wheels, but in the long run could offer a lot of flexibilities for us (e.g., unlock us from being unable to fix driver-related issues, see https://github.com/uber-go/nilaway/labels/driver).

yuxincs:
Since golangci-lint v1.57.0, it offers a new [Module Plugin System](https://golangci-lint.run/plugins/module-plugins/), which is a lot more robust and easier to use than the plain Go plugin system. This unlocks the possibility of running NilAway in golangci-lint as a private linter.

This PR implements the require interface for the module plugin system for golangci-lint and updates the README to list the instructions on how to run NilAway with golangci-lint.

With this, our official recommendation to run NilAway would be to either run it via golangci-lint or bazel/nogo, the standalone checker is only provided for evaluation purposes due to its limitations (lack of fact caching, lack of support for conditional build etc.).

This also closes #175 since it is no longer required.

We should switch our own linting task and CI to run NilAway via golangci-lint as well, but that will happen in follow-up PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,8 +64,7 @@ nilaway -include-pkgs="<YOUR_PKG_PREFIX>,<YOUR_PKG_PREFIX_2>" ./...
 
 ### golangci-lint (>= v1.57.0)
 
-NilAway, in its current form, still reports a fair number of false positives. This makes NilAway
-fail to be merged with [golangci-lint][golangci-lint] and be offered as a linter
+NilAway, in its current form, can report false positives. This unfortunately hinders its immediate merging in [golangci-lint][golangci-lint] and be offered as a linter
 (see [PR#4045][pr-4045]). Therefore, you need to build NilAway as a plugin to golangci-lint to be
 executed as a private linter. There are two plugin systems in golangci-lint, and it is much easier
 to use the [Module Plugin System][golangci-lint-module-plugin] (introduced since v1.57.0), and it
--- a/README.md
+++ b/README.md
@@ -33,13 +33,6 @@ NilAway is implemented using the standard [go/analysis][go-analysis], making it 
 drivers (i.e., [golangci-lint][golangci-lint], [nogo][nogo], or [running as a standalone checker][singlechecker]).
 
 > [!IMPORTANT]  
-> Due to the sophistication of the analyses that NilAway does, NilAway caches its findings about a particular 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -61,6 +62,65 @@ Then, run the linter by:
 nilaway -include-pkgs="<YOUR_PKG_PREFIX>,<YOUR_PKG_PREFIX_2>" ./...
 ```
 
+### golangci-lint (>= v1.57.0)
+
+NilAway, in its current form, still reports a fair number of false positives. This makes NilAway
+fail to be merged with [golangci-lint][golangci-lint] and be offered as a linter
```
sonalmahajan15, line 68:
```suggestion
NilAway, in its current form, can report false positives. This unfortunately hinders its immediate merging in [golangci-lint][golangci-lint] and be offered as a linter
```

sonalmahajan15 (approved):
Nice work! Enabling NilAway to be plugged into the golangci-lint system is great. Everything looks good to me, I just have a small rephrasing suggestion.

PR #236, opened 2024-04-25
Add w/o grouping run in golden test

sonalmahajan15:
This PR updates the golden test in CI to run under two modes: with and without grouping of similar errors.


tools/cmd/golden-test/main.go
```diff
@@ -55,7 +55,7 @@ type BranchResult struct {
 
 // Run runs the golden tests on the base branch and the test branch and writes the summary and
 // diff to the writer.
-func Run(writer io.Writer, baseBranch, testBranch string) error {
+func Run(writer io.Writer, baseBranch, testBranch, groupingFlag string) error {
```
yuxincs, line 58: nit: it would probably be better if we actually take a bool here and convert to string within this function for better type safety :)

.github/workflows/golden-test.yml
```diff
@@ -9,8 +9,16 @@ on:
...
+    name: Golden Test ${{ matrix.grouping }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - grouping: 'w/ grouping'
+            group_errors: 'true'
+          - grouping: 'w/o grouping'
+            group_errors: 'false'
+
```
yuxincs, line 21: Creating two jobs (that are running in parallel) is a good idea since golden test can easily run for several minutes.

But this would actually create two comments in the PR, which may hinder the readability for other comments.

Instead, can we keep the two jobs here (with id `golden-test-w-grouping` and `golden-test-wo-grouping`) [with output](https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs), and extract the uploading step to be a separate job that depends on these two jobs, reads the output from these two jobs, and construct the message together and upload only once.

It shouldn't be too complex I think, and offers better messages for readability. What do you think?

tools/cmd/golden-test/main.go
```diff
@@ -178,12 +178,16 @@ func ParseDiagnostics(reader io.Reader) (map[Diagnostic]bool, error) {
 
 // WriteDiff writes the summary and the diff (if the base and test are different) between the base
 // and test diagnostics to the writer. If the writer is os.Stdout, it will write the diff in color.
-func WriteDiff(writer io.Writer, branches [2]*BranchResult) {
+func WriteDiff(writer io.Writer, branches [2]*BranchResult, groupingFlag string) {
```
yuxincs, line 181: same as above: should we just use a `bool` and only convert whenever we have to for better type safety?

yuxincs (approved):

PR #239, opened 2024-05-04
Refactor the contract analyzer

yuxincs:
This PR simply refactors the contract analyzer to prepare for further updates. Specifically, it 

(1) Defines a separate type for `[]Contract` such that it can be made to be exported (i.e., satisfies the `analysis.Fact` interface) in the future.
(2) Replaces a few places where we were returning initialized empty slices with simple `nil`. In principle we should not distinguish nil slice vs empty nonnil slice.
(3) Shortens a few variable names to help improve readability without introducing ambiguity (since the variables are all under the `functioncontracts` package already).
(4) Moved test cases from `src/go.uber.org/functioncontracts/xxx` to `src/go.uber.org/xxx` inside `functioncontracts` package. Those tests are only meant to be used for `functioncontracts` package, such that an extra folder only increases nesting but does not help test organization.

Note that this PR does not bring any functionality changes.


sonalmahajan15 (approved):
Looks good. The refactors in this PR make sense to me.

PR #241, opened 2024-05-13
Fix integration tests to not group error messages

yuxincs:
This PR disables the error messages when running integration tests of NilAway (against different analyzer drivers) such that the output is more accurate for comparisons.

As part of this change, we fixed one want string in testdata where one false positive was silently ignored due to grouping. It is now properly marked as FP and a TODO is added to remove it once cross-package contract support is fixed in NilAway.


sonalmahajan15 (approved):
Ah, this was a good catch!

PR #242, opened 2024-05-14
Remove util.TypeOf helper function

yuxincs:
This PR simply removes the helper function `util.TypeOf`. Although it shortens the call from `pass.TypesInfo.TypeOf`, it increases the cognitive load such that we do not know which we should use in NilAway's codebase and we don't know if `util.TypeOf` has any special handling (it doesn't!). Moreover, we're using a mixture of these two functionally-identical expressions in our codebase.

Removal of such a helper function reduces confusions and helps maintainability.


sonalmahajan15 (approved):
LGTM!

PR #243, opened 2024-05-14
Remove ErrorOnNilableMap handling

yuxincs:
Reading from a nil map does not really panic and is actually used quite frequently in Go. When NilAway was initially developed we added a feature that errs when a nilable map is accessed, and guard it under an internal feature flag for us to see more data points. Now that NilAway is more mature and we have seen more real world examples, this PR removes such handling and the stale feature flag.


sonalmahajan15 (approved):
Agree with the cleanup!

PR #244, opened 2024-05-15
Fix convergence problem due to unbounded growing of assertion tree

sonalmahajan15:
This PR fixes the convergence issues causes by unbounded growing of assertion trees with assignments in a loop, such as in the simple example shown below.
```
func test() {
         a := &A{}
	for {
		a = a.f
	}
}
```
Here, the assertion tree should look like `root -> varAssertionNode (a) -> fldAssertionNode (f)`. However, the assertion before was growing unboundedly (`root -> a -> f -> f -> f -> ...`) until the `stableRoundLimit` was hit. This simple function was taking 7 iterations to converge, while after the code change it now takes only 3 iterations to converge.

Taking this opportunity, I have also added tests for testing fixpoint convergence. For this, I refactored existing infrastructure for anonymous functions and generalized it.

Note: the goal of this PR is to improve performance only, and should not have any effect on the reported errors.


yuxincs (approved):
Nice find and great refactor for test helpers!

PR #245, opened 2024-05-15
Fix for nested non-builtin call expression in map access

sonalmahajan15:
This PR refines the logic of non-literal map access, particularly in the case of built-in call expressions, to not consider the index trackable if there exists a nested call to non-trackable expression (e.g., custom function) from inside the built-in function.

For example, 
```
type MessageBlock struct{}

func (m *MessageBlock) Messages() []*int {
	return []*int{new(int)}
}

func test(msgSet []*MessageBlock) {
	for _, msgBlock := range msgSet {
		_ = *msgBlock.Messages()[len(msgBlock.Messages())-1]
	}
}
```
the `test` function was taking 7 iterations to converge, while now after the fix it requires only 3 iterations.

Note: the goal of this PR is to improve performance only, and should not have any effect on the reported errors.


yuxincs (approved):
Great catch!

PR #249, opened 2024-05-23
Refactor CFG preprocess logic

yuxincs:
This PR refactors the CFG preprocess logic. Specifically, we have extracted the rich check effect construction out of the preprocessing logic such that it is now in its standalone file. This offers a lot better modularity for the already-enormous `assertiontree` package.

Now in `BackpropAcrossFunc` function we are calling preprocess and rich-check-effect-generating functions separately.

### DIFF
--- a/assertion/function/assertiontree/preprocess_blocks.go
+++ b/assertion/function/assertiontree/preprocess_blocks.go
@@ -89,19 +89,6 @@ func copyGraph(graph *cfg.CFG) *cfg.CFG {
 	return newGraph
 }
 
-// stripNoops returns a copy of the passed slice `effects`, minus any no-ops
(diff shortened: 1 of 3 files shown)


assertion/function/assertiontree/preprocess_blocks.go
```diff
@@ -91,61 +89,6 @@ func copyGraph(graph *cfg.CFG) *cfg.CFG {
 	return newGraph
 }
 
-// stripNoops returns a copy of the passed slice `effects`, minus any no-ops
```
yuxincs, line 94: Code starting here is simply moved to `rich_check_effect.go`, no logic is really changed

PR #251, opened 2024-05-24
Refine the handling of pointers in `TypeAsDeepType`

sonalmahajan15:
This PR refines the handling of `*types.Pointer` in `TypeAsDeepType` to only consider a pointer type deep if its underlying elements are of a deep type. The implication of always considering a pointer deep was that for type like `var x *int; foo(x)`, unnecessary deep triggers were being created. The change in this PR reduces the creation of these redundant triggers.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -63,8 +63,9 @@ func TypeAsDeepType(t types.Type) (types.Type, bool) {
 	case *types.Chan:
 		return t.Elem(), true
 	case *types.Pointer:
-		// Only consider pointers to deep types (e.g., `var x *[]int`) as deep type, not pointers to basic types (e.g., `var x *int`)
-		if _, ok := t.Elem().(*types.Basic); !ok {
+		// Only consider pointers to deep types (e.g., `var x *[]int`) as deep type,
+		// not pointers to basic types (e.g., `var x *int`) or struct types (e.g., `var x *S`)
+		if _, ok := t.Elem().(*types.Basic); !ok && TypeAsDeeplyStruct(t.Elem()) == nil {
 			return t.Elem(), true
 		}
 	}
--- a/util/util.go
+++ b/util/util.go
@@ -63,10 +63,9 @@ func TypeAsDeepType(t types.Type) (types.Type, bool) {
 	case *types.Chan:
 		return t.Elem(), true
 	case *types.Pointer:
-		if _, ok := t.Elem().(*types.Basic); ok {
-			return nil, false
+		if _, ok := t.Elem().(*types.Basic); !ok {
+			return t.Elem(), true
 		}
-		return t.Elem(), true
 	}
 	return nil, false
 }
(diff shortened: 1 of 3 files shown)


util/util.go
```diff
@@ -57,7 +57,10 @@ func TypeAsDeepType(t types.Type) (types.Type, bool) {
 	case *types.Chan:
 		return t.Elem(), true
 	case *types.Pointer:
-		return t.Elem(), true
+		// Only consider pointers to deep types (e.g., `var x *[]int`) as deep type, not pointers to basic types (e.g., `var x *int`)
+		if _, ok := t.Elem().(*types.Basic); !ok {
+			return t.Elem(), true
+		}
```
yuxincs, line 63: Would this actually work for cases where it's a pointer to struct?

Do we actually want to call `TypeAsDeepType` recursively here? `return TypeAsDeepType(t.Elem())`?

yuxincs (commented):
Nice find!! I just have a quick question :)

util/util.go
sonalmahajan15: Ah yes, good point. Now added support for pointer to struct. Added more test cases too, to cover this and a named deep type case.

I actually did do it the recursive way first, but quickly realized that if there exists a circular named type dependency then this goes into an infinite loop. But moreover, we were missing deep type cases like pointer of pointer (`**x`), causing a change in number of errors as well. The hypothesis of this PR was to only remove the redundant triggers without affecting the number of errors, hence backtracked from the recursive approach.

yuxincs: > exists a circular named type dependency then this goes into an infinite loop

Hmmm, is it something like this?

```go
type A *B
type B *A
```

It's definitely weird, and the compiler doesn't complain (it complains if you do `type A B; type B A` though...).

(it would probably be good to add the pattern to the test suite to avoid accidental regressions)

> But moreover, we were missing deep type cases like pointer of pointer (**x), causing a change in number of errors as well.

Are you suggesting pointer of a named type, where the named type is also a pointer (e.g., `type A *B; type B *int`)? I think by default double pointers aren't allowed. I think other helper functions probably suffer from the same double-ptr problem, and we definitely don't need to handle them in this PR :) Also, are we missing cases like named types, i.e., `type A []int` where `A` should be considered deep as well, right?

Another proposal for simpler handling:

Instead of recursions, should we unwrap the ptr first, and then check it's type (like other utility functions in this package), e.g., 

```go
switch UnwrapPtr(t).(type) {
case ....
// no case for `*types.Pointer`
}
```

This definitely would miss the cases discussed above:
(1) double-ptrs: we're only unwrapping the ptr once, but it avoids infinite loops (and handling double ptrs should be more complex and in separate PRs).
(2) named types: well, if we do `UnwrapPtr(t.Underlying())`, we should be handling it fine?

sonalmahajan15: We had discussed this point extensively internally. Updating the logic accordingly, incorporating your suggestion and internal discussion.

yuxincs (approved):

PR #252, opened 2024-05-30
Add more test cases for calling no-return functions

yuxincs:
This PR adds more test cases to test NilAway's ability to handle no-return functions (such as `os.Exit`, `runtime.Goexit`, `testing.T.SkipNow` etc.).


assertion/function/preprocess/cfg.go
```diff
@@ -119,6 +124,31 @@ func copyGraph(graph *cfg.CFG) *cfg.CFG {
...
+		stmt, ok := node.(*ast.ExprStmt)
+		if !ok {
+			continue
+		}
+		call, ok := stmt.X.(*ast.CallExpr)
+		if !ok {
+			continue
+		}
+		if trustedfunc.TrimSuccsOn(p.pass, call) {
+			block.Succs = nil
```
sonalmahajan15, line 146: I was just wondering if we should set successors to nil or point to the return block... What is the behavior of the drivers where this behavior is handled correctly? Perhaps we can simulate that behavior here.

sonalmahajan15 (commented):
By and large looks good to me. Just a small question about whether we should set successors to nil or point to the return block.

yuxincs:
Actually, instead of doing hacky postfixes in NilAway, we should just fix `ctrlflow` analyzer: https://github.com/golang/tools/pull/497

Will repurpose this PR to just introduce those valuable test cases instead :)

assertion/function/preprocess/cfg.go
yuxincs: I compared the correct behaviors and it seems that `block.Succs == nil` is the behavior, so we're following that practice.

This is also documented: https://pkg.go.dev/golang.org/x/tools/go/cfg

> A block may have 0-2 successors: zero for a return block or a block that calls a function such as panic that never returns; one for a normal (jump) block; and two for a conditional (if) block.

Anyway, I think this comment is no longer applicable since we decided to just fix the upstream `ctrlflow` analyzer instead of doing hacky postfixes in NilAway.

sonalmahajan15: Got it. Makes sense. Thanks for clarifying :)

sonalmahajan15 (approved):
> Actually, instead of doing hacky postfixes in NilAway, we should just fix ctrlflow analyzer: https://github.com/golang/tools/pull/497
>
> Will repurpose this PR to just introduce those valuable test cases instead :)

Oh yeah, makes complete sense! :)

PR #254, opened 2024-06-05
Fix for making rich check effect handling unbiased for return statement ordering

sonalmahajan15:
This PR fixes the bug which made rich check effect handling sensitive to return statement ordering, resulting in false positives for certain cases. The effect of ordering of statements is illustrated in the example below.

```
// Problematic
func retPtrErr() (*int, error) {
	if cond {
		return new(int), retNil()
	}
	return nil, retErr()
}

// Works fine
func retPtrErr() (*int, error) {
	if cond {
	       return nil, retErr()	
	}	
        return new(int), retNil()
}
```

The problem is with no entry being created in `inferredMap` for non-nil error return from `retErr()` in `StoreImplication`, since we don't propagate non-nilness forward. This creates a problem when the anonymous function parameter in `FilterTriggersForErrorReturn` analyzes for nilability of return sites, where it incorrectly assumes that absence in `inferredMap` implies unknown nilability. This PR corrects that logic.

### DIFF
--- a/inference/engine.go
+++ b/inference/engine.go
@@ -262,22 +262,24 @@ func (e *Engine) ObservePackage(pkgFullTriggers []annotation.FullTrigger) {
 				isDeep := kind == annotation.DeepConditional
 				primitive := e.primitive.site(site, isDeep)
 				if val, ok := e.inferredMap.Load(primitive); ok {
-					switch vType := val.(type) {
-					case *DeterminedVal:
+					if vType, ok := val.(*DeterminedVal); ok {
 						if !vType.Bool.Val() {
 							return assertiontree.ProducerIsNonNil
 						}
-					case *UndeterminedVal:
-						// Consider the producer site as non-nil, if the determined value is non-nil, i.e.,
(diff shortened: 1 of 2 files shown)


inference/engine.go
```diff
@@ -220,7 +222,7 @@ func (e *Engine) ObservePackage(pkgFullTriggers []annotation.FullTrigger) {
 	filteredTriggers := nonErrRetTriggers
 	// Remove deleted triggers from nonErrRetTriggers (if needed).
 	if len(delTriggers) != 0 {
-		filteredTriggers = make([]annotation.FullTrigger, 0, len(nonErrRetTriggers)-len(delTriggers))
```
sonalmahajan15, line 223: Subtraction of `len(delTriggers)` removed from capacity setting, since now with the changed logic `delTriggers` can sometimes contain deletion of error returning triggers which are not in `nonErrRetTriggers`, causing `len(delTriggers) > len(nonErrRetTriggers)`.

yuxincs (approved):

PR #264, opened 2024-07-03
Support for "always safe" for error returning and ok-form functions

sonalmahajan15:
This PR adds support for "always safe" case for rich check effect functions, namely error returning functions and ok-form functions.

The support for always safe tracking is currently limited to: 

1. directly determinable cases of "non-nil" return
```
      (a) return &S{}, errors.New(...)  // supported
      (b) return new(S), errors.New(...)  // supported
      (c) if s != nil {
               return s, errors.New(...)  // supported
          }
  
      (d) return getS(), errors.New(...)  // not supported
      (e) return x.y.z.f, errors.New(...)   // not supported
```

2. immediate function call only
```
func errRetFunc() (*S, error) {
      return &S{}, errors.New(...)
}

func wrapErrRetFunc() (*S, error) {
      return errRetFunc()
}

func test1() {
       // supported
       v, _ := errRetFunc()    // immediate function call, supported
       _ = *v

       // not supported
       v, _ := wrapErrRetFunc()  // multiple hops, not supported
       _ = *v
}
```

### DIFF
--- a/assertion/function/assertiontree/root_assertion_node.go
+++ b/assertion/function/assertiontree/root_assertion_node.go
@@ -269,8 +269,7 @@ func (r *RootAssertionNode) AddConsumption(consumer *annotation.ConsumeTrigger) 
 					Consumer: consumer,
 				})
 			}
-
-			return // expr is not trackable, but cannot be nil, so do nothing
+			return
 		}
 		if len(producers) != 1 {
 			panic("multiply-returning function call was passed to AddConsumption")
--- a/assertion/function/assertiontree/root_assertion_node.go
+++ b/assertion/function/assertiontree/root_assertion_node.go
@@ -258,7 +258,9 @@ func (r *RootAssertionNode) AddConsumption(consumer *annotation.ConsumeTrigger) 
 	path, producers := r.ParseExprAsProducer(consumer.Expr, false)
 	if path == nil { // expr is not trackable
 		if producers == nil {
-			if _, ok := consumer.Annotation.(*annotation.UseAsReturnForAlwaysSafePath); ok {
+			// Here we can infer that the expression is non-nil by definition. Instead of ignoring creation of a trigger,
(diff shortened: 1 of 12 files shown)


inference/engine.go
```diff
@@ -154,18 +180,65 @@ func (e *Engine) ObserveAnnotations(pkgAnnotations *annotation.ObservedMap, mode
...
+
+	// Filter out the triggers that are to be deleted.
+	var filteredPkgFullTriggers []annotation.FullTrigger
+	for i, t := range pkgFullTriggers {
+		if triggersToBeDeleted[i] {
+			continue
+		}
+		filteredPkgFullTriggers = append(filteredPkgFullTriggers, t)
+	}
+	pkgFullTriggers = filteredPkgFullTriggers
```
yuxincs, line 227: Use [slices.DeleteFunc](https://pkg.go.dev/slices#DeleteFunc) to modify the slice in-place?

assertion/function/assertiontree/root_assertion_node.go
```diff
@@ -258,6 +258,18 @@ func (r *RootAssertionNode) AddConsumption(consumer *annotation.ConsumeTrigger)
...
+				r.AddNewTriggers(annotation.FullTrigger{
+					Producer: &annotation.ProduceTrigger{
+						Annotation: &annotation.ProduceTriggerNever{},
+						Expr:       consumer.Expr,
+					},
+					Consumer: consumer,
+				})
+			}
+
 			return // expr is not trackable, but cannot be nil, so do nothing
```
yuxincs, line 273: remove this comment?

yuxincs (approved):

inference/engine.go
sonalmahajan15: Good idea. Done :)

assertion/function/assertiontree/root_assertion_node.go
sonalmahajan15: Done

PR #265, opened 2024-07-06
Better support json format

madneal:
This PR is aimed to provide a better suport for the json format.

## Current Issue
 
For the orginal json flag provided by [checker.go](https://github.com/golang/tools/blob/0e7ccc0478c322a6b32fb9f74ebfc42e4f965ee9/go/analysis/internal/checker/checker.go#L510), it's impossible to get `analysisflags.JSON` from nilaway. There is a conflict bettwen the `pretty-print` and `json` output currently due to the code https://github.com/uber-go/nilaway/blob/e90288479601315af13b7fdd3ccd6b50c53a8e7c/nilaway.go#L43.
As the default `conf.PrettyPrint ` is true, so the output will be messy when try to output as json.
image

## Pros
* To import the `format` flag, it's possible to implement the output format by nilaway itself.
* It's possible to implement more format without the limits of [tools](https://github.com/golang/tools)

Can use this flag like this:
```
nilaway -format="json" ./...
```


madneal:
@yuxincs can u help to review this?

madneal:
there are still some issues for this PR, I will try to fix them at first.

PR #266, opened 2024-07-13
add tip for the JSON output

madneal:
This PR provides a tip when output as JSON due to the current default `pretty-print` flag

## Current Issue
 
For the orginal json flag provided by [checker.go](https://github.com/golang/tools/blob/0e7ccc0478c322a6b32fb9f74ebfc42e4f965ee9/go/analysis/internal/checker/checker.go#L510), it's impossible to get `analysisflags.JSON` from nilaway. There is a conflict bettwen the `pretty-print` and `json` output currently due to the code https://github.com/uber-go/nilaway/blob/e90288479601315af13b7fdd3ccd6b50c53a8e7c/nilaway.go#L43.
As the default `conf.PrettyPrint ` is true, so the output will be messy when try to output as json.
image


sonalmahajan15 (approved):
Awesome, we had missed realizing this problem. Thanks for helping us improve the documentation!

madneal:
@sonalmahajan15 seems some issue with the golden test

sonalmahajan15:
@madneal , yes, something seems to be wrong. We'll look into it and merge the PR.

yuxincs:
Yeah, I think we'll have to wait for https://github.com/golang/go/issues/61324 to have an importable checker API to use.

yuxincs:
I'm updating your branch since we recently fixed the CI issue, and we'll need those changes in order for CI to pass :)

yuxincs:
Thanks @madneal for the contributions!

Issue #269, opened 2024-08-14
panic with go1.23.0


jayalane:
I just added nilaway and got my build clean (thanks, very nice code; I'm adding more devs to my project, so need to be less trusting of my callers :)

Went to upgrade to 1.23.0 next but the nilaway binary is failing on my repo:aa

```
rlcs(1:595)$ go install go.uber.org/nilaway/cmd/nilaway@latest
Tue Aug 13 17:55:37 PDT 2024a
/Users/chlane/go/src/github.paypal.com/chlane/rlproxyserv/rlcs
Tue Aug 13 17:55:37 PDT 2024
rlcs(1:596)$ nilaway ./...
panic: Cannot range over: func(yield func(E) bool)

goroutine 9900 [running]:
golang.org/x/tools/go/ssa.(*builder).rangeStmt(0x14039cece40, 0x14039cf8680, 0x14003b64420, 0x0)
	/Users/chlane/go/pkg/mod/golang.org/x/tools@v0.18.0/go/ssa/builder.go:2281 +0x698
golang.org/x/tools/go/ssa.(*builder).stmt(0x14039cece40, 0x14039cf8680, {0x10538dad0?, 0x14003b64420?})
	/Users/chlane/go/pkg/mod/golang.org/x/tools@v0.18.0/go/ssa/builder.go:2505 +0x1ac
golang.org/x/tools/go/ssa.(*builder).stmtList(...)
	/Users/chlane/go/pkg/mod/golang.org/x/tools@v0.18.0/go/ssa/builder.go:910
golang.org/x/tools/go/ssa.(*builder).stmt(0x14039cece40, 0x14039cf8680, {0x10538d950?, 0x140041f68d0?})
	/Users/chlane/go/pkg/mod/golang.org/x/tools@v0.18.0/go/ssa/builder.go:2467 +0xbf0
golang.org/x/tools/go/ssa.(*builder).buildFromSyntax(0x14039cece40, 0x14039cf8680)
	/Users/chlane/go/pkg/mod/golang.org/x/tools@v0.18.0/go/ssa/builder.go:2579 +0x204
golang.org/x/tools/go/ssa.(*builder).buildFunction(0x10530e401?, 0x14039cf8680)
	/Users/chlane/go/pkg/mod/golang.org/x/tools@v0.18.0/go/ssa/builder.go:2533 +0x110
golang.org/x/tools/go/ssa.(*builder).iterate(0x14039cece40)
[...]

PR #270, opened 2024-08-14
new x/tools dep to work with newer go1.23.0

jayalane:
As per issue https://github.com/uber-go/nilaway/issues/269 this PR updates go.mod/go.sum to work with go 1.23.0


arnested:
I experience the same problem and this fixes it for me as well.

chlanejayasinha:
Signed.  (I had signed it at the time but I put in a different email that I thought was associated to the account but wasn't; just now I signed it with an email that was associated to the account).

chlanejayasinha:
The failing test looks like it succeeded but then failed to connect back to api.github.com with the results.  403.  a

yuxincs (approved):

yuxincs:
Thanks @jayalane!

PR #271, opened 2024-08-15
Fix golden test permission issue

yuxincs:


yuxincs:
So it seems to be only triggered by PRs from other repositories. Maybe it's an permission issue. Will create another PR from a forked repo instead.

PR #273, opened 2024-08-20
Fix golden test for forked repository

yuxincs:
We were facing permission issues for `GITHUB_TOKEN` in our Golden Test CI job for forked repository. This is because the `GITHUB_TOKEN` by default does not have permissions to post comments by design for security reasons.

The recommended way from GitHub is to break this into two parts: (1) the first job that executes the (potentially malicious) code from forked repository with limited default permission, and upload the results as a non-executable artifact, and (2) the second job that is triggered by the completion of the first job via `workflow_run` trigger. This job always runs on main branch, and has proper permissions to post comments. It downloads the artifact from (1) and posts the comment.


yuxincs:
We did not see the Golden Test comment because the 2nd job has to be in main branch in order for this to take effect. I have run it in my forked repository as an example https://github.com/yuxincs/nilaway/pull/2

sonalmahajan15 (approved):
This is awesome! Glad you fixed the problem for our OSS contributors!

PR #278, opened 2024-09-13
Generalize trusted function framework to be a hook framework

yuxincs:
This PR generalizes the trusted function framework to be a NilAway-wide hook framework.

To be honest it was already a NilAway-wide thing: it hooks into different parts of the system and can do different actions based on certain conditions. This PR simply lifts it up to the root level and call it "hook" instead of trusted function: in the future we may leverage this framework to handle more than function calls, with more types of actions, and hooking into more points (e.g., even in inference).

This PR is the first of a series of PRs to generalize this framework: this one simply does the relocation with no changes other than reference updates.

### DIFF
--- a/assertion/function/assertiontree/backprop_util.go
+++ b/assertion/function/assertiontree/backprop_util.go
@@ -23,7 +23,7 @@ import (
 	"go/types"
 
 	"go.uber.org/nilaway/annotation"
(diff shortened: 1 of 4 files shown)


sonalmahajan15 (approved):
Like the refactoring!

PR #280, opened 2024-09-16
Break splitBlocksOn into a new hook

yuxincs:
Similar to #279, this PR breaks the logic that splits the CFG block upon seeing a trusted function into a separate strongly-typed hook for better maintainability and slightly better performance (due to fewer lookups in hook points). It introduces no functionality changes.


assertion/function/preprocess/cfg.go
```diff
@@ -263,7 +257,7 @@ func (p *Preprocessor) restructureConditional(graph *cfg.CFG, thisBlock *cfg.Blo
 				newCond := &ast.BinaryExpr{
 					X:     x,
 					Y:     y,
-					Op:    token.EQL, // As discussed, we change the operator to EQL here.
+					Op:    token.EQL, // SplitBlockOn discussed, we change the operator to EQL here.
```
sonalmahajan15, line 260:
```suggestion
					Op:    token.EQL, // As discussed, we change the operator to EQL here.
```

sonalmahajan15 (approved):
Looks good. Small comment about typo correction.

PR #281, opened 2024-09-16
Add modeling for `errors.As` in the generalized hook framework

### LINKED ISSUE #95: Add support for `errors.As` in trusted functions framework
`error.As` reports a false positive as shown below. Add support for [errors.As](https://pkg.go.dev/errors#As), and other similar functions in the trusted functions framework.

```
func retErr() error {
	return nil
}

type myErr struct{}

func (myErr) Error() string { return "myErr message" }

func test() {
	err := retErr()
	var x *myErr
	if errors.As(err, &x) {
		print(*x) // error reported here
	}
}
```

yuxincs:
This PR introduces a new hook point in the CFG preprocessing logic to replace conditionals with equivalent expressions (along with a few simple renames and comment additions for better readability).

With the new hook, we are introducing modeling for `errors.As(err, &target)` since it is equivalent to `errors.As(err, &target) && target != nil`. This makes the implication explicit such that NilAway is able to understand it during the analysis.

Note that technically `target` can still be nil even if `errors.As(err, &target)` is true. For example, if err is a typed nil (e.g., `var err *exec.ExitError`), then `errors.As` would actually find a match, but `target` would be set to the typed nil value, resulting in a `nil` target. However, in practice this should rarely happen such that even the official documentation assumes the target is non-nil after such check [1]. So here we make this assumption as well.

[1] https://pkg.go.dev/errors#As

Fixes #95

### DIFF
--- a/assertion/function/preprocess/cfg.go
+++ b/assertion/function/preprocess/cfg.go
@@ -55,7 +55,14 @@ func (p *Preprocessor) CFG(graph *cfg.CFG, funcDecl *ast.FuncDecl) *cfg.CFG {
 	failureBlock := &cfg.Block{Index: int32(len(graph.Blocks))}
 	graph.Blocks = append(graph.Blocks, failureBlock)
 
-	// Perform the (series of) CFG transformations.
+	// Perform a series of CFG transformations here (for hooks and canonicalization). The order of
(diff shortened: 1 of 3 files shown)


sonalmahajan15 (approved):
Nice! Like the support for `errors.As()` 👍

PR #283, opened 2024-09-16
Remove dead code for custom nilable types

yuxincs:
We do not really have custom nilable types for a really long time now, so it makes sense to remove the dead code.

If we ever want to add such custom nilable types again, we should consider adding a hook here via the hook framework (#278).


sonalmahajan15 (approved):
Agree with the code cleanup

PR #286, opened 2024-10-08
Handle newly-introduced types.Alias type

yuxincs:
Go 1.22 [1] introduces a proper `types.Alias` type for type aliases. This PR adds handling (by simply unwrapping alias node - effectively restoring previous behavior) for such a type.

Further simplification and refactors will be in future PRs.


sonalmahajan15 (changes requested):
Nice! Could you also add support for `iter.Seq`? The golden test is reporting internal errors.

yuxincs:
> ```diff
> AppendSeq
> ```

This seems to be a separate issue we should handle, will do that in a follow-up PR :)

yuxincs:
> > Nice! Could you also add support for `iter.Seq`? The golden test is reporting internal errors.
> 
> This seems to be a separate issue we should handle, will do that in a follow-up PR :)

Done in #288

sonalmahajan15 (approved):
Awesome. Good to land after the needed support added in #288 .

Issue #287, opened 2024-10-10
Handle iterators introduced in Go 1.23


yuxincs:
Go 1.23 introduced [iterators](https://go.dev/wiki/RangefuncExperiment), which NilAway does not support yet. This issue tracks such support. 

Initially, we should at least assume these are deeply nonnil without panicking.

spacewander:
Is there someone still working on this issue? Currently, nilaway can't work with programs written with Go 1.23.

yuxincs:
Hey @spacewander , could you share the logs for "nilaway can't work with programs written with Go 1.23"? 

We had a PR landed to at least not panicking for iterators, and we assume the results from the iterators are always non-nil (such that it can lead to false negatives), but it should be working fine for other patterns.

spacewander:
Glad to hear that there is an ongoing PR addresses this issue!
Sorry for my previous words. In my situation, an APP has been migrated to Go 1.23, and after introducing the iterators feature, the nilaway won't work anymore. Would give a try once the PR is merged. Could you tell me which PR is solving this issue so that I could subscribe to it?

yuxincs:
It is actually already merged #288, could you verify that you're on the latest version?

Also, when you say "won't work anymore", could you clarify it a bit? Does it show some errors (panics), or does it simply stop reporting previously-reported diagnostics? 🤔

spacewander:
I got some messages about iter when running nilaway with a project configured with Go 1.23:

```
.../pkg/mod/golang.org/toolchain@v0.0.1-go1.23.2.darwin-arm64/src/slices/iter.go:50:17: cannot range over seq (variable of type iter.Seq[E])
.../pkg/mod/golang.org/toolchain@v0.0.1-go1.23.2.darwin-arm64/src/maps/iter.go:51:20: cannot range over seq (variable of type iter.Seq2[K, V])
.../pkg/mod/golang.org/x/net@v0.33.0/html/iter.go:50:17: cannot range over n.ChildNodes() (value of type iter.Seq[*Node])
```

Not sure if it is related.

yuxincs:
Hmmmm, we should have support for `iter.Seq` and `iter.Seq2` now. Not sure why it didn't work for you. It also looks like a toolchain issue, and we have recently upgraded the dependencies (#303).

While investigating we have also found out that we are not handling iterators in the general sense (they can be `func(func () bool)`, `func(func(K) bool)`, or `func(func(K, V) bool)`). PR #304 should handle it.

[...]

PR #288, opened 2024-10-10
Add initial handling for iterators

yuxincs:
NilAway is panicking on code that ranges over iterators. This PR adds initial handling for such that we assume the results from the iterators are always nonnil, without panicking.

See #287


sonalmahajan15 (approved):
LGTM!

PR #293, opened 2024-11-16
feat: allow zero slice using func

### LINKED ISSUE #268: Refine tracking of slicing index
For the below case, NilAway correctly tracks the implication of slicing an empty slice and **_does not_** report an error.
```
var x []int
_ = x[:0]
```

However, when the slicing index is a non-literal, like below, NilAway currently reports a false positive. Refine tracking of non-literal indices to understand such code patterns and avoid false positives.
```
var x []int
_ = x[:min(len(x), 100)] // FP: unassigned variable `x` sliced into
```

fatelei:
fix #268

### DIFF
--- a/assertion/function/assertiontree/root_assertion_node.go
+++ b/assertion/function/assertiontree/root_assertion_node.go
@@ -1191,9 +1191,22 @@ func (r *RootAssertionNode) isType(expr ast.Expr) bool {
 // even when the slice itself is nil, i.e, one of [:0] [0:0] [0:] [:] [:0:0] [0:0:0]
 func (r *RootAssertionNode) isZeroSlicing(expr *ast.SliceExpr) bool {
 	l, h, m := expr.Low, expr.High, expr.Max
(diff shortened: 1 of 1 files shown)


ccoVeille:
No unit test? 😱🤔

fatelei:
> No unit test? 😱🤔

i haven't find where to add ut

ccoVeille:
Let's wait for a maintainer.

But good thing you took this issue

ccoVeille:
As it's about slice, I would say here maybe

https://github.com/uber-go/nilaway/blob/main/testdata%2Fsrc%2Fgo.uber.org%2Fslices%2Fslices.go

fatelei:
> As it's about slice, I would say here maybe
> 
> https://github.com/uber-go/nilaway/blob/main/testdata%2Fsrc%2Fgo.uber.org%2Fslices%2Fslices.go

ok, i will check it

yuxincs:
Closing for now, but feel free to reopen and continue the discussion!

PR #302, opened 2025-03-07
Heuristic for error wrapping functions

sonalmahajan15:
This PR adds a heuristic for identifying error wrapping functions (e.g., `errors.Wrapf(err, "message")`), which are presumed to return a non-nil error value. A function is deemed as an "error wrapping function" if:
- it has at least one argument of error-implementing type.
- it has at least one return value of error-implementing type.

The purpose of this heuristic is to be able to determine the error wrapping function from its signature only, even when the body of the function may not be accessible.


sonalmahajan15:
I will raise this PR again since I would like to test it with coderabbit

sonalmahajan15:
@coderabbitai review

PR #305, opened 2025-03-07
Handle basic type alias in range statement

yuxincs:
We had handling for `range <basic_type>` (e.g., `range 10`). However, if the `<basic_type>` is type alias or a defined type where the underlying type is basic, NilAway actually crashes.

This PR fixes that and prettifies the format a little bit for readability.

### DIFF
--- a/assertion/function/assertiontree/backprop.go
+++ b/assertion/function/assertiontree/backprop.go
@@ -457,31 +457,30 @@ func backpropAcrossRange(rootNode *RootAssertionNode, lhs []ast.Expr, rhs ast.Ex
 			produceAsDeepRHS(1) // If we are not ranging over a string, then we cannot assume basic type
 		}
 	case 1:
-		if util.TypeIsDeeplyMap(rhsType) ||
-			util.TypeIsDeeplySlice(rhsType) ||
-			util.TypeIsDeeplyArray(rhsType) ||
-			typeIsString(rhsType) {
-			produceAsIndex(0) // If we are ranging over a map slice or string with only a single
-			// lhs operand, then that operand will be int-valued
+		// If we are ranging over a map slice or string with only a single lhs operand, then that
+		// operand will be int-valued.
+		if util.TypeIsDeeplyMap(rhsType) || util.TypeIsDeeplySlice(rhsType) || util.TypeIsDeeplyArray(rhsType) || typeIsString(rhsType) {
+			produceAsIndex(0)
 			return nil
 		}
+		// Iterating over a channel with only a single lhs operand will still result in deeply
+		// produced lhs values.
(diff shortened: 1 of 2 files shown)


assertion/function/assertiontree/backprop.go
```diff
@@ -457,31 +457,30 @@ func backpropAcrossRange(rootNode *RootAssertionNode, lhs []ast.Expr, rhs ast.Ex
...
-			// still result in deeply produced lhs values
+			produceAsDeepRHS(0)
 			return nil
 		}
 
 		// Here the range is over basic types, such as integers (e.g., "for i := range 10").
 		// We do not need to do anything here, as the basic types are always presumed to be non-nil.
-		if _, ok := rhsType.(*types.Basic); ok {
+		if _, ok := rhsType.Underlying().(*types.Basic); ok {
 			return nil
```
yuxincs, line 476: I'm actually getting confused by the codebase now, but perhaps @sonalmahajan15 can help?

Here in some conditions we are explicitly calling `produceNonnil`, which basically adds `TriggerNever` trigger, but in some other cases we are simply not adding any triggers. Are there any subtle differences that I'm not remembering?

assertion/function/assertiontree/backprop.go
```diff
@@ -457,31 +457,30 @@ func backpropAcrossRange(rootNode *RootAssertionNode, lhs []ast.Expr, rhs ast.Ex
...
-			produceAsDeepRHS(0) // iterating over a channel with only a single lhs operand will
-			// still result in deeply produced lhs values
+			produceAsDeepRHS(0)
 			return nil
 		}
 
 		// Here the range is over basic types, such as integers (e.g., "for i := range 10").
 		// We do not need to do anything here, as the basic types are always presumed to be non-nil.
-		if _, ok := rhsType.(*types.Basic); ok {
+		if _, ok := rhsType.Underlying().(*types.Basic); ok {
```
yuxincs, line 475: BTW this is the only change, all other changes are simply formatting chanegs

sonalmahajan15:
@coderabbitai review

assertion/function/assertiontree/backprop.go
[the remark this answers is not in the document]
yuxincs: I'm confused about this comment here, is there a potential issue? the following text basically says the change here _fixes_ the issue.

sonalmahajan15: IIRC explicit `TriggerNever` producer is created if there is a potential for a consumer to be present, while in other cases such as Basic type, where by nilability doesn't apply by design, not trigger is added.

sonalmahajan15 (approved):

Issue #91, opened 2023-11-02
Support cross-package nolint parsing


yuxincs:
We can report errors on `pkg A` when we are analyzing `pkg B` due to our multi-pkg inference. However, nogo’s support for nolint comments is based on reading the AST from current package.

That is, nogo does not know we have a nolint comment in pkg A when analyzing pkg B , but NilAway can report an error there, and hence cannot be suppressed.

The other linter implementations most likely follow the same logic.

We’ll need to implement own cross-pkg nolint support (i.e., reading the nolint and writing it to facts, then read it in downstream analysis to decide whether to report a certain error or not).

davecb:
I think you're constrained to use the "//nolint:" notation, but _just to clarify my understanding_, am I correct in thinking what you mean it to express is "this cannot be nil by construction"?

I just ran nilaway on a program that of mine, and added `// postcondition: x is non-nil` to remind me that the library code I just called produces a non-nil result. (I sort of miss #define ASSERT(...) from a less-modern language (;-))

yuxincs:
@davecb Not exactly :)

`//nolint:` here basically just suppresses errors reported, without meaning anything else.

You're probably thinking about adding annotations to parameters & returns, similar to what Java has (`@Nullable` and `@Nonnull`). 

In NilAway we try to avoid adding annotation burden onto devs since it's not a builtin language feature unlike Java. But we do parse comments like 

```
//nonnil(result 0)
func foo() (*int) {....}
```

Where the return value would be _assumed_ nonnil in NilAway's perspective. But those are mostly for testing purposes (do not rely on this feature in your code, though).

In the future, we plan to add a tiny "annotation"-like library that NilAway understands, _and_ most importantly will log and/or panic if the assumption is broken.

```
func main() {
  // say for some complex code in foo NilAway just cannot reason the return value is nonnil, but you as a developer know that for sure.
  v := foo() 
  v := castToNonnil(v) 
  print(*v) // NilAway won't complain here anymore.
}
```

where `castToNonnil` does nothing but return the same value, in the meantime it can log and/or panic if the input is ever `nil` (i.e., breaks the assumption) to (1) combat false positives in NilAway in edge cases, and (2) still raise awareness of code defects.

yuxincs:
This is now fixed by #320

PR #320, opened 2025-04-11
Support cross-package nolint suppressions

### LINKED ISSUE #316: Add support for //nolint:nilaway
Add support for //nolint:nilaway.

```
//nolint:nilaway
func testme() {
	var x *int
	_ = *x  // report no error here
}

func callTestme() {
	testme()
}
```

yuxincs:
NilAway _may_ report diagnostics for upstream packages when seeing more evidence of nil flows in current package. 

For example, if there is a function `upstream.Foo(p *int)` that simply dereferences the given pointer. When analyzing the upstream package NilAway does not know whether that would constitute a potential nil panic by itself -- there is not a nil flow from nilable site to the dereference site *yet*.

Then in current package NilAway sees a `upstream.Foo(nil)` call, it will connect the dots and report a diagnostic on the dereference site inside `upstream.Foo(p *int)`. Note that this happens *while analyzing the current package*.

If users have nolint comment set on the upstream package, such suppression will not be respected by analyzer drivers since the implementations focus on reading the nolint comments within the current package and suppress the diagnostics reported for current package.

We can wait for drivers to pick up such support, but since NilAway can be/is used by different drivers, offering this support directly in NilAway would lead to better user experience (consistency between drivers) and easier maintenance (we implement it once in NilAway).

This PR does exactly that: we read the nolint comments related to NilAway, and export the scopes of those suppressions as `Fact`s. When NilAway is ready to report diagnostics in the `diagnostic.Engine`, we use all nolint suppressions (including the imported ones from upstream) and suppress there.

We have added test cases in our unit tests (for package-local testing) and integration tests (for cross-package testing).

Fixes #316 #91

## Summary by CodeRabbit

- **New Features**
  - Added support for recognizing and processing `//nolint:nilaway` comments, allowing suppression of lint warnings for nil pointer dereferences, including across package boundaries.
  - Enhanced diagnostic capabilities to honor `nolint` directives at line, function, and file levels.

- **Tests**
  - Introduced comprehensive test cases to validate local and cross-package suppression of nil pointer dereference warnings using various forms of `nolint` comments.

### DIFF
--- a/accumulation/analyzer.go
+++ b/accumulation/analyzer.go
@@ -45,7 +45,7 @@ var Analyzer = &analysis.Analyzer{
 	Doc:        _doc,
 	Run:        run,
 	FactTypes:  []analysis.Fact{new(inference.InferredMap)},
-	Requires:   []*analysis.Analyzer{config.Analyzer, assertion.Analyzer, annotation.Analyzer},
+	Requires:   []*analysis.Analyzer{config.Analyzer, assertion.Analyzer, annotation.Analyzer, diagnostic.NoLintAnalyzer},
 	ResultType: reflect.TypeOf(([]analysis.Diagnostic)(nil)),
 }
 
--- a/diagnostic/conflict.go
+++ b/diagnostic/conflict.go
@@ -18,10 +18,10 @@ import (
 	"fmt"
 	"go/ast"
 	"go/token"
-	"path/filepath"
 	"strings"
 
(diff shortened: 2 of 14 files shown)


sonalmahajan15 (approved):
Awesome work! :)

vijaykramesh:
see https://github.com/uber-go/nilaway/issues/321 @yuxincs but this broke our CI

PR #323, opened 2025-04-18
Upgrade golangci-lint to v2

yuxincs:
This PR upgrades golangci-lint we use to v2, where the configs are migrated. We also applied the automated fixes from the linters due to the upgrade.


assertion/function/assertiontree/root_assertion_node.go
```diff
@@ -569,13 +569,14 @@ func (r *RootAssertionNode) AddComputation(expr ast.Expr) {
 		//
 		// A similar approach is followed for the `||` operator, where we only need to care about the false branch since the
 		// Y expression won't be executed if the X expression is true.
-		if expr.Op == token.LAND {
+		switch expr.Op {
+		case token.LAND:
```
sonalmahajan15, line 573: Same as above: change to switch intended or LLM introduced?

annotation/produce_trigger.go
```diff
@@ -460,7 +460,7 @@ func (BlankVarReturnPrestring) String() string {
 // DuplicateParamProducer duplicates a given produce trigger, assuming the given produce trigger
 // is of FuncParam.
 func DuplicateParamProducer(t *ProduceTrigger, location token.Position) *ProduceTrigger {
-	key := t.Annotation.(*FuncParam).TriggerIfNilable.Ann.(*ParamAnnotationKey)
+	key := t.Annotation.(*FuncParam).Ann.(*ParamAnnotationKey)
```
sonalmahajan15, line 463: Was this change intended? Or is it only LLM introduced? I am fine with these changes being clubbed with this PR, but was just curious. :)

sonalmahajan15 (approved):

PR #324, opened 2025-04-18
Add golangci-lint driver to the integration tests

yuxincs:
Currently our integration tests only test the standalone driver for cross-package inference. This PR also adds golangci-lint driver to the test suite to catch regressions. Note that we use the same version as the one used to lint NilAway's codebase. This way we only have a single gcl version in our codebase for easier maintenance.

On the high level, the golangci-lint integration test driver does the following:

1. Creates temp dir to host the custom configs and binary that will be built.
2. Generates `.custom-gcl.yaml` config from template `.custom-gcl.template.yaml` in `testdata/integration` and store it in the temp dir.
3. Install a bootstrap golangci-lint binary, this will be used to build the custom-gcl.
4. Build the custom golangci-lint binary (`custom-gcl`) in the temp dir.
5. Runs `custom-gcl` on `testdata/integration`
6. Processes results into position-diagnostic map

Then it goes back to the integration test logic to compare against the `want` strings.


sonalmahajan15 (approved):
Nice work!

PR #326, opened 2025-05-16
Support for named anonymous error return functions

sonalmahajan15:
This PR adds support for error return functions defined as named anonymous functions. Example:
```
func test() {
	f := func() (*int, error) {
		if dummy {
			return nil, &myErr2{}
		}
		return new(int), nil
	}

	x, err := f()
	if err != nil {
		return
	}
	_ = *x  // safe dereference
}
```
The PR makes primarily 3 changes:
- track error returning names anonymous functions in `rich_check_effect.go`.
- add its corresponding producer handling in `parse_producer_expr.go` guarded by the `EnableAnonymousFunc` flag. Note that in the absence of the flag, the PR employs a temporary fix to suppress the errors, which means that we don't report false positives, but we also don't report true positives, leading to false negatives. 
- refactors and genealizes `FuncIsErrReturning` to accept `*types.Signature`, instead of `*types.Func`.

### DIFF
--- a/assertion/function/assertiontree/rich_check_effect.go
+++ b/assertion/function/assertiontree/rich_check_effect.go
@@ -345,7 +345,7 @@ func NodeTriggersOkRead(rootNode *RootAssertionNode, nonceGenerator *util.GuardN
 
 		rhsFuncDecl, ok := rootNode.ObjectOf(callIdent).(*types.Func)
 
-		if !ok || !util.FuncIsOkReturning(rhsFuncDecl) {
+		if !ok || !util.FuncIsOkReturning(rhsFuncDecl.Signature()) {
 			return nil, false
 		}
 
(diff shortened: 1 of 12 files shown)


assertion/function/assertiontree/rich_check_effect.go
```diff
@@ -397,15 +397,30 @@ func NodeTriggersFuncErrRet(rootNode *RootAssertionNode, nonceGenerator *util.Gu
...
+	switch t2 := obj.(type) {
+	case *types.Func:
+		sig = t2.Signature()
+	case *types.Var:
+		if anonFunc, ok := t2.Type().(*types.Signature); ok {
+			sig = anonFunc
+		}
+	default:
+		return nil, false
+	}
```
yuxincs, line 415: I wonder if we should have a helper function for this in `typeshelper` that extracts `*types.Signature` for both normal function or anonymous function (or function pointer), this could be used in many places I presume.

yuxincs (approved):

assertion/function/assertiontree/rich_check_effect.go
sonalmahajan15: Good point. Done.

PR #327, opened 2025-05-23
Disable error grouping for unit tests

yuxincs:
With error grouping for unit tests, some error messages could be grouped together. This could lead to problems where all errors are grouped into a single source:

```
var nilable *int

switch  {
case "case 1 should have an error":
case "case 2 should NOT have an error":
}
```

the error is grouped since they have the same nil source. However, if case 2 does indeed report an error, we will not be able to detect such discrepancy due to error grouping.

This PR disables the error grouping for unit test (and extracted a test package `errormessage` to `grouping` since it mainly tests the error messages w/ grouping enabled). 

With this, we found a bug where we were not handling `testing.TB.Fatal` correctly, which will be fixed in a follow-up PR.


testdata/src/go.uber.org/abnormalflow/abnormalflow.go
```diff
@@ -78,19 +78,19 @@ func testDirectDereference(msg string, t *testing.T, b *testing.B, f *testing.F,
...
-		print(*nilable)
+		print(*nilable) //want "unassigned variable `nilable` dereferenced"
 	case "testing.TB.Skip":
 		tb.Skip()
-		print(*nilable)
+		print(*nilable) //want "unassigned variable `nilable` dereferenced"
 	case "testing.TB.Skipf":
 		tb.Skipf("msg")
-		print(*nilable)
+		print(*nilable) //want "unassigned variable `nilable` dereferenced"
```
yuxincs, line 93: These errors shouldn't be here, but will be fixed in a follow-up PR

testdata/src/go.uber.org/abnormalflow/abnormalflow.go
```diff
@@ -235,26 +235,26 @@ func testErrReturn(msg string, val bool, t *testing.T, b *testing.B, f *testing.
...
 			tb.Skip(err)
 		}
-		print(*ptr)
+		print(*ptr) //want "dereferenced"
 	case "testing.TB.Skipf":
 		if err != nil {
 			tb.Skipf("msg %s", err)
 		}
-		print(*ptr)
+		print(*ptr) //want "dereferenced"
```
yuxincs, line 258: These errors shouldn't be here, but will be fixed in a follow-up PR

sonalmahajan15 (approved):
Ah, good catch.

PR #329, opened 2025-05-23
Move all stub files in test code together under stubs/

yuxincs:
This PR simply moves the stub files out of `trustedfuncs` specific test suite to be all under `stubs/` for easier management, such that it can be shared by all test code.

This also frees our hook regexes that we no longer need to match wild prefixes -- instead, we can just say "(stubs/)?github.com/...."


testdata/src/go.uber.org/trustedfunc/testing/testing.go
```diff
@@ -1,20 +0,0 @@
-//  Copyright (c) 2023 Uber Technologies, Inc.
```
yuxincs, line 1: This is actually not needed, we can just rely on stdlib `testing` library without stubs for it.

PR #330, opened 2025-05-25
Fix for handling `new` in error wrapping heuristic

sonalmahajan15:
This PR further refines the error wrapping heuristic by adding support for use of built-in `new` as argument. Example: `Wrap(new(wrapped), "some message")`.

### DIFF
--- a/hook/assume_return.go
+++ b/hook/assume_return.go
@@ -83,6 +83,7 @@ func isErrorWrapperFunc(pass *analysis.Pass, call *ast.CallExpr) bool {
 	if !ok {
 		return false
 	}
+
 	if util.FuncIsErrReturning(funcObj.Signature()) {
 		args := call.Args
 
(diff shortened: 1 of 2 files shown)


hook/assume_return.go
```diff
@@ -63,6 +63,22 @@ func isErrorWrapperFunc(pass *analysis.Pass, call *ast.CallExpr) bool {
 		return false
 	}
 
+	// If the call expr is built-in `new`, then we check if its argument type implements the error interface.
+	// This case particularly gets triggered for the expression: `Wrap(new(MyErrorStruct), "message")`.
+	if obj == util.BuiltinNew {
+		if len(call.Args) == 0 {
```
yuxincs, line 69: Does this ever happen (since here we know it is the builtin new)?

yuxincs (approved):

hook/assume_return.go
sonalmahajan15: You are right, this doesn't happen and can be safely removed. I had added it at some point when I think I was trying to generalize it beyond `new`, but no longer needed.

PR #332, opened 2025-05-29
Support for unnamed anonymous error returning functions

sonalmahajan15:
This PR extends the functionality of #326 by adding support for error return functions defined as unnamed anonymous functions. Example:

```
func test() {
	x, err := func() (*int, error) {
		if dummy {
			return nil, &myErr2{}
		}
		return new(int), nil
	}()
	if err != nil {
		return
	}
	_ = *x  // safe dereference
}
```

The PR makes primarily 3 changes:

- generalize `NodeTriggersFuncErrRet` in `rich_check_effect.go` to work on function signatures.
- update case of *ast.CallExpr in `parse_producer_expr.go` to handle `*ast.FuncLit`. It is guarded by the `EnableAnonymousFunc` flag. Note that in the absence of the flag, the PR employs a temporary fix to suppress the errors, which means that we don't report false positives, but we also don't report true positives, leading to false negatives.

### DIFF
--- a/assertion/function/assertiontree/parse_expr_producer.go
+++ b/assertion/function/assertiontree/parse_expr_producer.go
@@ -331,7 +331,7 @@ func (r *RootAssertionNode) ParseExprAsProducer(expr ast.Expr, doNotTrack bool) 
 			return nil, r.getFuncReturnProducers(fun.Sel, expr)
 
 		case *ast.FuncLit:
-			// TODO: This case can be possibly be combined with the case of *ast.Ident above.
+			// TODO: This case can possibly be combined with the case of *ast.Ident above.
 			var funcIdent *ast.Ident
 
 			if r.functionContext.functionConfig.EnableAnonymousFunc {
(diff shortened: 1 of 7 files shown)


assertion/function/assertiontree/parse_expr_producer.go
```diff
@@ -329,6 +330,37 @@ func (r *RootAssertionNode) ParseExprAsProducer(expr ast.Expr, doNotTrack bool)
 			// function call has non-literal args, so is not literal, use its return annotation
 			return nil, r.getFuncReturnProducers(fun.Sel, expr)
 
+		case *ast.FuncLit:
+			// TODO: This case can be possibly be combined with the case of *ast.Ident above.
```
yuxincs, line 334:
```suggestion
			// TODO: This case can possibly be combined with the case of *ast.Ident above.
```

assertion/function/assertiontree/rich_check_effect.go
```diff
@@ -390,20 +389,15 @@ func NodeTriggersFuncErrRet(rootNode *RootAssertionNode, nonceGenerator *util.Gu
...
+	// Get signature of the function call (normal and anonymous both)
+	var sig *types.Signature
+	tv := rootNode.Pass().TypesInfo.Types[callExpr.Fun]
+	if tv.Type != nil {
+		if s, ok := tv.Type.(*types.Signature); ok {
+			sig = s
+		}
 	}
 
-	sig := typeshelper.GetFuncSignature(obj)
```
yuxincs, line 406: why wouldn't `GetFuncSignature` work here? 🤔

yuxincs (approved):

PR #333, opened 2025-06-01
Recognize and suppress type alias false positives

sonalmahajan15:
This PR adds support for recognizing type alias, but employs a temporary fix of suppressing the false positives caused by type aliases until we implement a comprehensive reasoning support for the same.

Example of a type alias:
```

type MyErrRetFunc = func() (*int, error)

func callTypeAliasFunc(f MyErrRetFunc) {
	x, err := f()
	if err != nil {
		return
	}
	_ = *x  // False positive was earlier reported here
}
```

### DIFF
--- a/assertion/function/assertiontree/parse_expr_producer.go
+++ b/assertion/function/assertiontree/parse_expr_producer.go
@@ -276,16 +276,14 @@ func (r *RootAssertionNode) ParseExprAsProducer(expr ast.Expr, doNotTrack bool) 
 			// 		}
 			// 		_ = *x
 			// }
-			// TODO: this is only a temporary fix to suppress the case of type aliases for functions.
+			// TODO: this is only a temporary fix to suppress false positives caused by type aliases.
 			//  Remove this once we have implemented complete support for type aliases.
 			t := r.Pass().TypesInfo.TypeOf(fun)
-			if t != nil {
-				if _, ok := t.(*types.Alias); ok {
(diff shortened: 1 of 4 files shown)


assertion/function/assertiontree/parse_expr_producer.go
```diff
@@ -266,6 +267,25 @@ func (r *RootAssertionNode) ParseExprAsProducer(expr ast.Expr, doNotTrack bool)
...
+			// 		x, err := f()
+			// 		if err != nil {
+			// 			return
+			// 		}
+			// 		_ = *x
+			// }
+			// TODO: this is only a temporary fix to suppress false positives caused by type aliases.
+			//  Remove this once we have implemented complete support for type aliases.
+			t := r.Pass().TypesInfo.TypeOf(fun)
+			if _, ok := t.(*types.Alias); ok {
```
yuxincs, line 282: Is there a reason we can't simply unwrap the type alias by `types.Unalias` or `.Underlying()`?

yuxincs:
LGTM but I have a quick question on if we can simply just unwrap the type here

assertion/function/assertiontree/parse_expr_producer.go
sonalmahajan15: Hmm, good question. I did not try this way. Let me check if it's possible.

sonalmahajan15: Discussed offline and decided to pursue this later.

PR #344, opened 2025-06-27
Add NilAway support for analyzing templ files

yuxincs:
This PR adds initial support for https://github.com/a-h/templ generated Go files.

Specifically, templ-generated files contain heavy uses of function literals, where NilAway does not have mature-enough support for it (it is still experimental). To stop the gap temporarily, we "inline" the function literals.

We have also added a templ example to the integration test suite to prevent future regressions.


util/asthelper/asthelper_test.go
```diff
@@ -69,7 +67,7 @@ func TestDocContains(t *testing.T) {
 		tc := tc
 		t.Run(tc.description, func(t *testing.T) {
 			t.Parallel()
-			require.Equal(t, tc.want, DocContains(tc.node, tc.s))
+			// require.Equal(t, tc.want, DocContains(tc.node, tc.s))
```
sonalmahajan15, line 70: Left commented by mistake?

sonalmahajan15 (approved):
Looks great!

PR #345, opened 2025-06-29
Fix for handling assignment from non-pointer to pointer type in rich check effect context

sonalmahajan15:
This PR addresses false positive due to lack of support for non-pointer to pointer types in rich check effect functions. Example:
```
type V struct{}

func retV() (V, error) {
   return V{}, errors.New("some error")
}

func test() {
   var v *V
   if a, err := retV(); err == nil {
        v = &a
        _ = *v  // False positive was reported here
   }
}
```

### DIFF
--- a/assertion/function/assertiontree/backprop.go
+++ b/assertion/function/assertiontree/backprop.go
@@ -312,10 +312,27 @@ func backpropAcrossAssignment(rootNode *RootAssertionNode, lhs, rhs []ast.Expr) 
 			return backpropAcrossTypeSwitch(rootNode, lhsIdent, r.X)
 		}
 
-		// Range statement of the form `for x := range y`, which is not overly complex to
-		// handle but does involve distinct semantics.
-		if r, ok := rhsNode.(*ast.UnaryExpr); ok && r.Op == token.RANGE {
-			return backpropAcrossRange(rootNode, lhs, r.X)
+		if r, ok := rhsNode.(*ast.UnaryExpr); ok {
+			switch r.Op {
+			case token.RANGE:
+				// Range statement of the form `for x := range y`, which is not overly complex to
+				// handle but does involve distinct semantics.
+				return backpropAcrossRange(rootNode, lhs, r.X)
+			case token.AND:
+				if !rootNode.functionContext.functionConfig.EnableStructInitCheck {
+					// This is the case of creating a pointer and assigning it to a variable, e.g., `x := &y`,
+					// where y is a non-pointer type (e.g., y := S{}).
+					// Here, the pointer is always nonnil, so we can just add a ProduceTriggerNever.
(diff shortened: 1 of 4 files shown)


assertion/function/assertiontree/backprop.go
```diff
@@ -312,10 +312,27 @@ func backpropAcrossAssignment(rootNode *RootAssertionNode, lhs, rhs []ast.Expr)
...
-		if r, ok := rhsNode.(*ast.UnaryExpr); ok && r.Op == token.RANGE {
-			return backpropAcrossRange(rootNode, lhs, r.X)
+		if r, ok := rhsNode.(*ast.UnaryExpr); ok {
+			switch r.Op {
+			case token.RANGE:
+				// Range statement of the form `for x := range y`, which is not overly complex to
+				// handle but does involve distinct semantics.
+				return backpropAcrossRange(rootNode, lhs, r.X)
+			case token.AND:
+				if !rootNode.functionContext.functionConfig.EnableStructInitCheck {
```
yuxincs, line 322: How does struct init handle this currently?

Asking because I'm thinking if we can just do this without the check (since the logic here is obviously correct)

assertion/function/assertiontree/rich_check_effect.go
```diff
@@ -354,8 +354,7 @@ func NodeTriggersOkRead(rootNode *RootAssertionNode, nonceGenerator *util.GuardN
 		for i := 0; i < len(lhs)-1; i++ {
 			lhsExpr := lhs[i]
 			lhsValueParsed := parseExpr(rootNode, lhsExpr)
-			if lhsValueParsed == nil || util.ExprBarsNilness(rootNode.Pass(), lhsExpr) {
-				// ignore assignments to any variables whose type bars nilness, such as 'int'
+			if lhsValueParsed == nil {
```
yuxincs, line 357: Hmmmm, if we now track all vars with types that are not subject to nilness, would it cause a big performance penalty?

yuxincs (approved):
Awesome! Just two questions, but otherwise logic LGTM (and love the added test cases)

assertion/function/assertiontree/backprop.go
sonalmahajan15: I know, I had thought of doing that as well. But struct init seems to have its own handling for this, because I was getting many missing errors when running the struct init test suite without the check. I did not spend time in coalescing the logic between struct init and otherwise, since we need to revisit and revamp struct init anyway.

assertion/function/assertiontree/rich_check_effect.go
sonalmahajan15: I don't think it should have any performance implications since this is only annotating rich check blocks, which does not create triggers. I'll run our internal performance evaluation script nonetheless to be sure.

PR #348, opened 2025-07-02
Allow complex length check conditions to determine the slice being nonnil

yuxincs:
We can have different length checks in Go code i.e., `0 < len(a) - 1` instead of simply `0 < len(a)` which would imply that `a` is nonnil. However, NilAway currently only reasons about simple cases like `0 < len(a)`. This PR adds the ability to reason about complex length check conditions by traversing the AST node and extracting the `len(a)` node.

Note that we maintain an unsound assumption that the length check yields nonnil `a` (unless explicitly checking for `len(a) == 0`) to reduce false positives.

### DIFF
--- a/assertion/function/assertiontree/util.go
+++ b/assertion/function/assertiontree/util.go
@@ -22,6 +22,7 @@ import (
 
 	"go.uber.org/nilaway/annotation"
 	"go.uber.org/nilaway/util"
+	"go.uber.org/nilaway/util/asthelper"
 	"golang.org/x/tools/go/analysis"
 	"golang.org/x/tools/go/ast/astutil"
 )
@@ -198,17 +199,6 @@ func AddNilCheck(pass *analysis.Pass, expr ast.Expr) (trueCheck, falseCheck Root
 		return noop, noop, true
 	}
 
-	asLenCall := func(expr ast.Expr) (ast.Expr, bool) {
-		if call, ok := expr.(*ast.CallExpr); ok {
-			if fun, ok := call.Fun.(*ast.Ident); ok {
-				if fun.Name == "len" && len(call.Args) == 1 {
-					return call.Args[0], true
-				}
-			}
-		}
-		return nil, false
-	}
-
 	isLiteralZeroInt := func(expr ast.Expr) bool {
 		if lit, ok := expr.(*ast.BasicLit); ok {
 			if lit.Kind == token.INT && lit.Value == "0" {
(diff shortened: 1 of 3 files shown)


assertion/function/assertiontree/backprop_util.go
```diff
@@ -826,10 +826,10 @@ func addAssignmentToConsumer(lhs, rhs ast.Expr, pass *analysishelper.EnhancedPas
 	var lhsExprStr, rhsExprStr string
 	var err error
 
-	if lhsExprStr, err = asthelper.PrintExpr(lhs, pass, true /* isShortenExpr */); err != nil {
+	if lhsExprStr, err = asthelper.PrintExpr(lhs, pass.Fset, true /* isShortenExpr */); err != nil {
 		return fmt.Errorf("converting LHS of assignment to string: %w", err)
 	}
-	if rhsExprStr, err = asthelper.PrintExpr(rhs, pass, true /* isShortenExpr */); err != nil {
+	if rhsExprStr, err = asthelper.PrintExpr(rhs, pass.Fset, true /* isShortenExpr */); err != nil {
```
yuxincs, line 832: Reduced the required fields to just `pass.Fset` to avoid circular dependency on `analysishelper`.

assertion/function/assertiontree/util.go
```diff
@@ -126,42 +127,6 @@ func detachFromParent(node AssertionNode, whichChild int) {
...
-	case token.GTR:
-		return token.LEQ
-	case token.LEQ:
-		return token.GTR
-	case token.GEQ:
-		return token.LSS
-	}
-	panic(fmt.Sprintf("unrecognized token %s has no known inverse", t))
-}
-
```
yuxincs, line 164: Moved to `tokenhelper`

assertion/function/assertiontree/util.go
```diff
@@ -199,171 +164,144 @@ func AddNilCheck(pass *analysishelper.EnhancedPass, expr ast.Expr) (trueCheck, f
...
 			},
 		},
 	}
 
-	// this applies each of the checkers to see if we can use it to trigger a return from this function
-	// the converse, inverse, and contrapositive of each exprCheck is checked as well
+	// Apply the checkers.
 	for _, check := range checkers {
-		if binExpr.Op == check.op {
+		// Note that `op` might be equal to `converse(op)` (e.g., `==` = `==`), so we must check
```
sonalmahajan15, line 286:
```suggestion
		// Note that `op` might be equal to `converse(op)` (e.g., `==` = `!=`), so we must check
```

I suppose you meant to show that EQ and NEQ are the same?

sonalmahajan15 (approved):

assertion/function/assertiontree/util.go
yuxincs: It's actually `==` (op) and `==` (converse of op). Maybe this is just too confusing, moved the example here :)

PR #369, opened 2025-10-16
Refine error wrapper heuristic

sonalmahajan15:
This PR refines the error wrapping heuristic to cover:
- return of new error from input string message (e.g., `return NewInternalServerError("some error message")`)
- nested error wrapping calls (e.g., `return Wrapf(Wrapf(Wrapf(NewError(err.Error()))))`)

### DIFF
--- a/hook/assume_return.go
+++ b/hook/assume_return.go
@@ -30,20 +30,19 @@ import (
 // functions that are not analyzed by NilAway. For example, "errors.New" is assumed to return a
 // nonnil value. If the given call expression does not match any known function, nil is returned.
 func AssumeReturn(pass *analysishelper.EnhancedPass, call *ast.CallExpr) *annotation.ProduceTrigger {
-	trigger, ok := matchTrustedFuncs(pass, call)
-	if ok {
+	if trigger := matchTrustedFuncs(pass, call); trigger != nil {
 		return trigger
 	}
 	return AssumeReturnForErrorWrapperFunc(pass, call)
 }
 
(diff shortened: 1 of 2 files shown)


hook/assume_return.go
```diff
@@ -22,20 +22,28 @@ import (
...
+		return trigger
+	}
+	return AssumeReturnForErrorWrapperFunc(pass, call)
+}
+
+func matchTrustedFuncs(pass *analysishelper.EnhancedPass, call *ast.CallExpr) (*annotation.ProduceTrigger, bool) {
 	for sig, act := range _assumeReturns {
 		if sig.match(pass, call) {
-			return act(call)
+			return act(call), true
```
yuxincs: `trigger`'s nilability implies this boolean value, right? shouldn't we just do that?

sonalmahajan15: Fair enough, done

util/util.go
```diff
@@ -547,3 +547,22 @@ func IdentOf(expr ast.Expr) *ast.Ident {
...
+}
+
+// implementsError checks if the given object implements the error interface. It also covers the case of
+// interfaces that embed the error interface.
+func implementsError(t types.Type) bool {
+	if ErrorInterface == nil {
+		return false
+	}
+	return types.Implements(t, ErrorInterface)
+}
```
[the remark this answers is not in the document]
sonalmahajan15: Actually I simplified the code even more, since I found an `ImplementsError()` function that I was able to refactor and use.

PR #377, opened 2025-12-05
Revised large func size calculation

sonalmahajan15:
This PR revises the large function size calculation to use CFG blocks instead of token/byte size.

### DIFF
--- a/tools/cmd/golden-test/main.go
+++ b/tools/cmd/golden-test/main.go
@@ -135,7 +135,6 @@ func Run(writer io.Writer, baseBranch, testBranch string) error {
 
 		// Run the built NilAway binary on the stdlib and parse the diagnostics.
 		var buf bytes.Buffer
-		// Time the execution of NilAway.
 		start := time.Now()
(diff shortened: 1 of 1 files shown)


assertion/function/analyzer_test.go
```diff
@@ -200,6 +200,78 @@ func TestBackpropFixpointConvergence(t *testing.T) {
 	}
 }
 
+// TestFunctionSizeLimit tests the function size limiting logic with various CFG block counts.
+func TestFunctionSizeLimit(t *testing.T) {
```
yuxincs: I actually feel the tests here are not testing anything (we are not even calling prod code).

Should we introduce a gigantic function (say, a switch with 500 case, can be easily generated by AI), and at the end introduce an obvious nil panic as test code (where NilAway won't report since we skipped it)?

yuxincs (approved):

assertion/function/analyzer_test.go
sonalmahajan15: Added large functions under `assertion/function/testdata/src/go.uber.org/funcsizelimit/largefunc.go` to test the new function size logic.

PR #382, opened 2025-12-23
Add option to show full filenames in output

### LINKED ISSUE #381: Filename truncation in error output complicates life for AI tools
$subj says it all, while nilaway is very useful for humans, I've found that AI tools struggle with the filenames in the output. I'd like to change this so that the path is printed relative to the current working directory rather than just the last two path components.

PR coming!

TvdW:
Fixes #381


TvdW:
Done: configurable, slightly more correct, and now with a test to avoid an accidental regression (since it's configurable)

yuxincs (approved):
Awesome, thanks for the contributions!! 🍻

PR #383, opened 2026-01-03
Fix panic on empty identifier in generic function calls

ewhauser:
Add missing IsEmptyExpr check when handling generic function calls to prevent panic when analyzing code like `_, err := genericFunc[T]()`.

The empty identifier `_` was being passed to ParseExprAsProducer which explicitly rejects it. This matches the pattern used in other similar code paths (lines 326, 365, 416).


ewhauser:
test added

PR #397, opened 2026-01-28
Model gRPC status.Error as always returning non-nil error

### LINKED ISSUE #367: False positive for status.Error with non-OK gRPC codes
The linter incorrectly flags a return nil, err statement where err is generated by google.golang.org/grpc/status.Error.

The linter seems to assume that status.Error() can return a nil error, which is only true if the gRPC code is codes.OK. When a non-OK code like codes.InvalidArgument is used, the function is guaranteed to return a non-nil error. The linter's static analysis doesn't appear to differentiate between these cases, leading to a false positive.

```golang
package main

import (
	"google.golang.org/grpc/codes"
	"google.golang.org/grpc/status"
)

// This function will be flagged by the nilerr linter.
func someFunction(isValid bool) (*string, error) {
	if !isValid {
		// The linter complains on the following line, even though
		// codes.InvalidArgument guarantees a non-nil error.
		return nil, status.Error(codes.InvalidArgument, "input is invalid")
	}
	result := "success"
	return &result, nil
}
```

The output will be:
```bash
literal `nil` returned from `someFunction()` in position 0 when the error return in position 1 is not guaranteed to be non-nil through all paths
```

Expected Behavior:
The linter should not report an error for this code, as status.Error(codes.InvalidArgument, "...") is guaranteed to produce a non-nil error, making a (nil, nil) return impossible.

nickita-khylkouski:
## Summary
Fixes #367

Model `google.golang.org/grpc/status.Error` as always returning a non-nil error.

## Problem
NilAway incorrectly flagged code like:
```go
return nil, status.Error(codes.InvalidArgument, "input is invalid")
```

The analyzer assumed `status.Error()` could return nil, which is only true when the code is `codes.OK` (a rare edge case). In practice, non-OK codes guarantee a non-nil error.

## Solution
Added `status.Error` to the library modeling in `hook/assume_return.go`, treating it as a non-nil producer following the same pattern as `strings.Split` and `errors.Join` (trading soundness for practicality on the rare OK-code edge case).

The fix handles:
- Direct function calls: `status.Error(codes.InvalidArgument, "...")`
- Package-level calls: `status.Error(codes.InvalidArgument, "...")`
- Both regular import and stub import paths

## Verification
- ✅ `make build` - successful
- ✅ `go test -short ./...` - all packages pass
- ✅ `make lint` - golangci-lint, mod tidy, and nilaway self-lint all pass
- ✅ `go vet ./...` - clean

## Trade-offs
Modeled as always returning non-nil for simplicity. The edge case where `status.Error(codes.OK, msg)` returns nil is rare in practice and out of scope for NilAway (consistent with `strings.Split` behavior where we assume always non-nil slice).

### DIFF
--- a/hook/assume_return.go
+++ b/hook/assume_return.go
@@ -190,7 +190,7 @@ var _assumeReturns = map[trustedFuncSig]assumeReturnAction{
 	{
 		kind:           _func,
 		enclosingRegex: regexp.MustCompile(`^(stubs/)?(google\.golang\.org/grpc|google\.golang\.org/grpc/status)$`),
-		funcNameRegex:  regexp.MustCompile(`^Error$`),
+		funcNameRegex:  regexp.MustCompile(`^Errorf?$`),
 	}: nonnilProducer,
 }
 
--- a/testdata/src/go.uber.org/trustedfunc/inference/grpc.go
+++ b/testdata/src/go.uber.org/trustedfunc/inference/grpc.go
@@ -1,0 +1,24 @@
+package inference
+
+import (
+	"stubs/google.golang.org/grpc/codes"
+	"stubs/google.golang.org/grpc/status"
+)
+
+func grpcStatusErrorTest(c string) (*int, error) {
+	switch c {
+	case "status.Error with non-OK code":
+		// status.Error with a non-OK code always returns non-nil error.
+		// NilAway should not flag this.
+		return nil, status.Error(codes.InvalidArgument, "input is invalid")
+	case "status.Error (false negative - OK code)":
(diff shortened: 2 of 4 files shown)


yuxincs:
@nickita-khylkouski Thanks for the contributions!

This looks good to me as an interim solution, since we should probably try to at least analyze the first argument and see if we can determine its value at compile time (since for the OK case it would return a nilable error). 

Can you add some test code to prevent future regressions? the google libraries can be stubbed and stored in `testdata/stubs`

yuxincs:
Awesome, thanks!

PR #399, opened 2026-01-28
Model json.Unmarshal and xml.Unmarshal as guaranteeing non-nil output parameter

### LINKED ISSUE #342: Common pattern for json.Unmarshall with nil map flagged as false positive
Code sample:
```golang
package main

import (
	"encoding/json"
	"fmt"
)

func main() {
	var outMap map[string]any
	if err := json.Unmarshal([]byte("{}"), &outMap); err != nil {
		println(fmt.Sprint("err: ", err))
		return
	}
	outMap["hello"] = "Hello"
}
```

Nilaway Output:
```
~/niltest/main.go:14:2: error: Potential nil panic detected. Observed nil flow from source to dereference point: 
	- niltest/main.go:14:2: unassigned variable `outMap` written to at an index
```

Contract for unmarshal is specified [here](https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/encoding/json/decode.go;l=97):
> To unmarshal a JSON object into a map, Unmarshal first establishes a map to use. If the map is nil, Unmarshal allocates a new map. Otherwise Unmarshal reuses the existing map, keeping existing entries. 

Detection can be avoided by preallocating an empty map in the code, but (I think) using the nil map here is common. Implementation code of unmarshal is [reflection heavy](https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/encoding/json/decode.go;l=644) for dealing with a nil map, so there may be limits on how clever the code can be to detect non-nillness of the field.

Overall, I was not sure if Nilaway has similar capabilities to Nullaway that allow you to specify the contract of a standard lib methods, but I was wondering if it would make sense to allowlist this method in some meaningful way (i.e. specifying that the output value will be non-nil when the error is nil)

nickita-khylkouski:
Fixes #342

## Problem

A common pattern when using `json.Unmarshal` with a nil map/slice was being flagged as a false positive:

```go
var outMap map[string]any
if err := json.Unmarshal([]byte("{}"), &outMap); err != nil {
    return
}
outMap["hello"] = "Hello"  // Previously flagged as error
```

When `json.Unmarshal` returns a nil error, the output parameter is guaranteed to be non-nil because the function allocates maps/slices when they're nil. This contract was not being modeled by NilAway.

## Solution

Following the pattern of `errors.As`, we now model this contract using the `ReplaceConditional` hook system. The call expression is replaced with:

```
json.Unmarshal(data, &v) && v != nil
```

This adds an implicit nil check that makes NilAway understand the output parameter is non-nil after a successful unmarshal (error == nil).

The same pattern applies to `xml.Unmarshal`, which has identical behavior.

## Changes

- Added `jsonUnmarshalAction` function in `hook/replace_conditional.go`
- Updated `ReplaceConditional` map to handle `encoding/json.Unmarshal` and `encoding/xml.Unmarshal`
- Both functions now guarantee non-nil output parameter when error is nil
- Added test cases in `testdata/src/go.uber.org/trustedfunc/json_unmarshal.go`

## Testing

- ✅ All linting checks pass
- ✅ Full test suite passes
- ✅ New test cases verify the fix works correctly

## Alternative Approaches Considered

1. **Custom RichCheckEffect:** Too complex for this use case, would require deep knowledge of the system
2. **Split blocks on Unmarshal:** Incorrect approach - this is for assertion-style functions, not conditional contracts
3. **Modify assume_return.go:** Not suitable - the contract is conditional (only when error is nil), not absolute

The `ReplaceConditional` approach is the simplest and most appropriate solution, following the established pattern from `errors.As`.

### DIFF
--- a/hook/replace_conditional.go
+++ b/hook/replace_conditional.go
@@ -69,10 +69,42 @@ var _errorAsAction replaceConditionalAction = func(_ *analysishelper.EnhancedPas
 	}
 }
 
+// _jsonUnmarshalAction replaces a call to `json.Unmarshal(data, &v)` or `xml.Unmarshal(data, &v)` with
+// `json.Unmarshal(data, &v) && v != nil`. This handles the contract where these functions allocate
+// maps/slices when they're nil, so when the error is nil, the output parameter is non-nil.
+// This is similar to errors.As in concept - we keep the call expression and add an implicit nil check.
+var _jsonUnmarshalAction replaceConditionalAction = func(_ *analysishelper.EnhancedPass, call *ast.CallExpr) ast.Expr {
+	if len(call.Args) != 2 {
+		return nil
+	}
+
+	unaryExpr, ok := call.Args[1].(*ast.UnaryExpr)
+	if !ok || unaryExpr.Op != token.AND {
+		return nil
+	}
+
+	return &ast.BinaryExpr{
+		X:     call,
+		Op:    token.LAND,
+		OpPos: call.Pos(),
+		Y:     newNilBinaryExpr(unaryExpr.X, token.NEQ),
+	}
+}
+
 var _replaceConditionals = map[trustedFuncSig]replaceConditionalAction{
 	{
 		kind:           _func,
 		enclosingRegex: regexp.MustCompile(`^errors$`),
 		funcNameRegex:  regexp.MustCompile(`^As$`),
 	}: _errorAsAction,
+
+	// `encoding/json.Unmarshal` and `encoding/xml.Unmarshal`
+	// When error is nil, the output parameter (second argument, a pointer to map/slice) is guaranteed to be non-nil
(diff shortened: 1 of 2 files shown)


yuxincs:
Thanks for the contributions!

This doesn't seem to be correct, errors.As returns a bool such that we can replace it with errors.As && target != nil. These are equivalent.

However, here the json.Marshal or xml.Marshal actually returns an error, and such replacement does not really make sense. 

We should actually replace the error check err != nil with err != nil && v ! = nil. This could be a bit tricky since the error check usually wouldn't be at the same node.

With this PR, we would basically always consider v to be non-nil, right? Even if err == nil.

Perhaps I missed something, happy to discuss further.

yuxincs:
Closing the PR for now, please reopen if my understanding is incorrect. Thx!

PR #412, opened 2026-03-20
fix: respect TERM=dumb and NO_COLOR for ANSI escape codes

### LINKED ISSUE #105: Output always contains ANSI escape codes regardless of how the TERM variable is set
Programs should not emit ANSI escape codes if TERM is set to dumb.

mvanhorn:
## Summary

Skips ANSI escape codes in `PrettyPrintErrorMessage` when `TERM=dumb` or `NO_COLOR` is set.

## Why this matters

`PrettyPrintErrorMessage` in `nilaway.go:61` unconditionally wraps error text with ANSI color codes (`\x1b[31m` for red, `\x1b[95m` for magenta, etc.). In terminals that don't support ANSI (like Emacs shell mode, some CI runners, or `TERM=dumb`), these appear as raw escape sequences in the output (#105).

## Changes

- `nilaway.go`: Added an early return in `PrettyPrintErrorMessage` that checks `os.Getenv("TERM") == "dumb"` and `os.Getenv("NO_COLOR") != ""` before applying ANSI codes. When either is set, returns plain text with "error: " prefix but no escape sequences. Also follows the [NO_COLOR convention](https://no-color.org/).

## Testing

`go build ./...` passes. The change is a 4-line guard at the top of an existing function. The existing `-pretty-print=false` flag is unaffected - it controls whether `PrettyPrintErrorMessage` is called at all (nilaway.go:47-48).

Fixes #105

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -19,12 +19,26 @@ import (
 	"flag"
 	"go/ast"
 	"go/types"
+	"os"
 	"reflect"
 	"strings"
 
 	"go.uber.org/nilaway/util/asthelper"
 	"golang.org/x/tools/go/analysis"
 )
+
+// defaultPrettyPrint returns the default for the PrettyPrint config option.
+// It respects the NO_COLOR and TERM=dumb conventions when no -pretty-print flag is provided.
+// https://no-color.org/ and POSIX terminal-capability conventions.
+func defaultPrettyPrint() bool {
+	if os.Getenv("NO_COLOR") != "" {
+		return false
+	}
+	if os.Getenv("TERM") == "dumb" {
+		return false
+	}
+	return true
+}
 
 // Config is the struct that stores the user-configurable options for NilAway.
 type Config struct {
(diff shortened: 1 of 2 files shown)


vincent067:
This looks like a good issue for a first-time contributor. I'd be happy to work on respecting TERM=dumb and NO_COLOR environment variables. Could you assign this to me?

mvanhorn:
Hey @vincent067 — this is actually the PR that addresses #105 (TERM=dumb / NO_COLOR support). It's been sitting waiting for maintainer review since March. If you're looking for a first-time contribution on nilaway, the issue tracker has a few open ones you could pick up. Review/testing on this PR would also be welcome if you want to dig in here.

yuxincs:
Hey, thanks for the contributions! 

Perhaps a better place to do this is at https://github.com/uber-go/nilaway/blob/main/config/config.go#L167 where we calculate the default value (currently it's always `true`, but we can make that respect the terminal envs)? This way if user wants to override it by giving the `-pretty-print` flag it would still be effective

yuxincs (approved):

mvanhorn:
Thanks @yuxincs, that is a much better placement. Pushed 234b235:

- Added `defaultPrettyPrint()` helper in `config/config.go` that returns false when `NO_COLOR` is set or `TERM=dumb`
- Used it as the default for the `-pretty-print` flag (around line 165)
- Reverted the env check from `PrettyPrintErrorMessage` in `nilaway.go`

The override path through `pass.Analyzer.Flags.Lookup(PrettyPrintFlag)` is untouched, so explicit `-pretty-print=true` still wins. Tests pass locally; lint fetch failed in my sandbox (network), so CI golangci-lint will be the source of truth there.

PR #413, opened 2026-04-15
Add `github.com/cockroachdb/errors` to hooks as alternative `errors` package

ian-oneleet:
https://github.com/cockroachdb/errors is a popular alternative to the stdlib `errors` package. It provides alternative versions of `errors.New`, `fmt.Errorf` (which it calls `errors.Newf`), and `errors.Join`, among other functions.

Nilaway currently flags more false positives for codebases that use `cockroachdb/errors` instead of stdlib `errors`, because the built-in special cases for stdlib `errors` don't carry over to `cockroachdb/errors`. This PR duplicates the special cases in `hook/assume_return.go` and `hook/replace_conditional.go` so that they work with `cockroachdb/errors`.

Code was produced with LLM assistance (Claude).

### DIFF
--- a/hook/assume_return.go
+++ b/hook/assume_return.go
@@ -161,12 +161,7 @@ var _assumeReturns = map[trustedFuncSig]assumeReturnAction{
 	{
 		kind:           _func,
 		enclosingRegex: regexp.MustCompile(`^(stubs/)?github\.com/cockroachdb/errors$`),
-		funcNameRegex:  regexp.MustCompile(`^New$`),
-	}: nonnilProducer,
-	{
-		kind:           _func,
-		enclosingRegex: regexp.MustCompile(`^(stubs/)?github\.com/cockroachdb/errors$`),
-		funcNameRegex:  regexp.MustCompile(`^Newf$`),
+		funcNameRegex:  regexp.MustCompile(`^New(f)?$`),
 	}: nonnilProducer,
 
 	// `errors.Join`
--- a/hook/assume_return.go
+++ b/hook/assume_return.go
@@ -157,6 +157,18 @@ var _assumeReturns = map[trustedFuncSig]assumeReturnAction{
 		funcNameRegex:  regexp.MustCompile(`^New$`),
 	}: nonnilProducer,
 
+	// `github.com/cockroachdb/errors`
+	{
+		kind:           _func,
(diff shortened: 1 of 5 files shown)


ian-oneleet:
@yuxincs friendly ping now that I've got CLA sign-off here 🙂

hook/assume_return.go
```diff
@@ -157,6 +157,18 @@ var _assumeReturns = map[trustedFuncSig]assumeReturnAction{
...
+	// `github.com/cockroachdb/errors`
+	{
+		kind:           _func,
+		enclosingRegex: regexp.MustCompile(`^(stubs/)?github\.com/cockroachdb/errors$`),
+		funcNameRegex:  regexp.MustCompile(`^New$`),
+	}: nonnilProducer,
+	{
+		kind:           _func,
+		enclosingRegex: regexp.MustCompile(`^(stubs/)?github\.com/cockroachdb/errors$`),
+		funcNameRegex:  regexp.MustCompile(`^Newf$`),
```
yuxincs: We can probably merge this with the above with a `(f)?`

testdata/src/stubs/github.com/cockroachdb/errors/errors.go
```diff
@@ -0,0 +1,25 @@
...
+package errors
+
+// nilable(err, target)
+func As(err error, target any) bool { return false }
+
+// nilable(result 0)
+func New(text string) error { return nil }
+
+// nilable(errs[], result 0)
+func Join(errs ...error) error { return nil }
```
yuxincs: These two actually return nonnil error per contract, could you update the stubs and annotations here?

yuxincs:
@ian-oneleet Hey, thanks for the contributions! I made two minor comments, could you take a look and see if they make sense? they should be simple fixes. After that I'll merge :)

hook/assume_return.go
ian-oneleet: Merged.

testdata/src/stubs/github.com/cockroachdb/errors/errors.go
ian-oneleet: I think I've corrected this, but not positive I did it right. Would appreciate another look!

yuxincs (approved):

PR #415, opened 2026-06-09
Improve zero slicing check by validating length bounds of the slice

### LINKED ISSUE #268: Refine tracking of slicing index

karthikbhata97:
This PR generalizes the zero-slicing check to a broader safe slicing check that also recognizes length-bounded indices.

NilAway previously suppressed nil-panic consumer triggers only for zero slicing — forms like b[:0], b[0:0], b[:], b[:0:0], b[0:0:0] where every present index is a compile-time zero. These are safe even when b is nil.

This missed a safe common pattern: slicing bounded by the slice's own length, e.g. b[:len(b)] or b[:min(len(b), 100)]. Since a nil slice has len == 0, any index bounded by len(b) evaluates to zero when b is nil, so the slicing cannot panic. NilAway reported false positives for these (#268).

### Changes

- Replace isZeroSlicing with isSafeSlicing, which validates each present low/high/max index independently via isSafeSliceIndex.
- An index is safe if it is absent, a compile-time zero, or length-bounded (isLengthBoundedSliceIndex).
- isLengthBoundedSliceIndex recognizes, recursively:
  - len(slice) — where the argument resolves to the same slice being sliced.
  - min(...) — length-bounded if at least one argument is length-bounded.
  - max(...) — length-bounded only if every argument is length-bounded.
- Add BuiltinMin / BuiltinMax to typeshelper.

cap(...) is intentionally not recognized as length-bounded; it is treated conservatively and still produces a consumer trigger.

### Testing

New testdata cases in slices.go covering:
- Length-bounded slicing that should not create a consumer (len, min, max, three-index forms, and parenthesized variants).
- Indices that should still create a consumer: len/cap of a different slice, cap of the sliced slice, and max(...) with an argument that is not length-bounded.

Fixes #268

### DIFF
--- a/assertion/function/assertiontree/root_assertion_node.go
+++ b/assertion/function/assertiontree/root_assertion_node.go
@@ -1229,16 +1229,16 @@ func (r *RootAssertionNode) isSafeSlicing(expr *ast.SliceExpr) bool {
 // isSafeSliceIndex returns true if `index` is safe to use as a slicing index for the potentially
 // nil slice `slice`, meaning it provably evaluates to zero when `slice` is nil. An absent index
 // (nil) and a compile-time zero are always safe, as is an index that is provably bounded by the
-// length of `slice` (see isLengthBounded).
+// length of `slice` (see isLengthBoundedSliceIndex).
 func (r *RootAssertionNode) isSafeSliceIndex(index, slice ast.Expr) bool {
-	return index == nil || r.Pass().IsZero(index) || r.isLengthBounded(index, slice)
+	return index == nil || r.Pass().IsZero(index) || r.isLengthBoundedSliceIndex(index, slice)
 }
 
-// isLengthBounded returns true if `index` is provably bounded by the length of `slice`, i.e.,
-// `index <= len(slice)`. Since a nil slice has len == 0, such an index evaluates to zero when the
-// slice is nil, making the slicing safe. The recognized patterns are `len(slice)`, `min(...)` where
-// at least one argument is length-bounded, and `max(...)` where every argument is length-bounded.
-func (r *RootAssertionNode) isLengthBounded(index, slice ast.Expr) bool {
+// isLengthBoundedSliceIndex returns true if `index` is provably bounded by the length of `slice`,
+// i.e., `index <= len(slice)`. Since a nil slice has len == 0, such an index evaluates to zero when
+// the slice is nil, making the slicing safe. The recognized patterns are `len(slice)`, `min(...)`
+// where at least one argument is length-bounded, and `max(...)` where every argument is length-bounded.
+func (r *RootAssertionNode) isLengthBoundedSliceIndex(index, slice ast.Expr) bool {
 	call, ok := ast.Unparen(index).(*ast.CallExpr)
 	if !ok {
 		return false
@@ -1255,14 +1255,14 @@ func (r *RootAssertionNode) isLengthBounded(index, slice ast.Expr) bool {
 	case typeshelper.BuiltinMin:
 		// `min(...)` is length-bounded if at least one of its arguments is.
 		for _, arg := range call.Args {
-			if r.isLengthBounded(arg, slice) {
+			if r.isLengthBoundedSliceIndex(arg, slice) {
 				return true
 			}
 		}
 	case typeshelper.BuiltinMax:
 		// `max(...)` is length-bounded only if all of its arguments are.
 		for _, arg := range call.Args {
-			if !r.isLengthBounded(arg, slice) {
+			if !r.isLengthBoundedSliceIndex(arg, slice) {
 				return false
 			}
 		}
--- a/assertion/function/assertiontree/root_assertion_node.go
+++ b/assertion/function/assertiontree/root_assertion_node.go
@@ -846,9 +846,10 @@ func (r *RootAssertionNode) AddComputation(expr ast.Expr) {
 	case *ast.SliceExpr:
 		// similar to index case
 
-		// zero slicing contains b[:0] b[0:0] b[0:] b[:] b[:0:0] b[0:0:0], which are safe even when b is
-		// nil, so we do not create consumer triggers for those slicing.
-		if !r.isZeroSlicing(expr) {
+		// safe slicing contains b[:0] b[0:0] b[0:] b[:] b[:0:0] b[0:0:0] and length-bounded forms such
+		// as b[:len(b)], which are safe even when b is nil, so we do not create consumer triggers for
+		// those slicing.
+		if !r.isSafeSlicing(expr) {
 			// For all the other slicing, the slice must be nonnil, so we create a consumer
 			// trigger.
 			r.AddConsumption(&annotation.ConsumeTrigger{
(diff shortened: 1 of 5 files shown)


assertion/function/assertiontree/root_assertion_node.go
```diff
@@ -1215,12 +1215,51 @@ func (r *RootAssertionNode) isType(expr ast.Expr) bool {
 }
 
 // isZeroSlicing returns if the given slice expression is a special case that will not cause panic
-// even when the slice itself is nil, i.e, one of [:0] [0:0] [0:] [:] [:0:0] [0:0:0]
+// even when the slice itself is nil. This holds when each of the present low, high, and max indices
+// provably evaluates to zero for a nil slice. Besides compile-time zero indices (e.g., [:0], [:0:0],
+// [0:0:0]), this also covers indices tied to the length or capacity of the slice (e.g., `x[:len(x)]`,
+// `x[:cap(x)]`, or `x[:min(len(x), 100)]`), since len(x) == cap(x) == 0 for a nil slice forces such
+// indices to zero.
 func (r *RootAssertionNode) isZeroSlicing(expr *ast.SliceExpr) bool {
```
yuxincs: Since we have expanded the support here, should we rename this to `isSafeSlicing`? it's no longer just checking if it's `isZeroSlicing`

assertion/function/assertiontree/root_assertion_node.go
```diff
@@ -1215,12 +1215,51 @@ func (r *RootAssertionNode) isType(expr ast.Expr) bool {
...
-	l, h, m := expr.Low, expr.High, expr.Max
-	return ((l == nil || r.Pass().IsZero(l)) && r.Pass().IsZero(h) && m == nil) || // [:0] [0:0]
-		((l == nil || r.Pass().IsZero(l)) && h == nil && m == nil) || // [0:] [:]
-		((l == nil || r.Pass().IsZero(l)) && r.Pass().IsZero(h) && r.Pass().IsZero(m)) // [:0:0] [0:0:0]
+	return r.isNilSafeSliceIndex(expr.Low, expr.X) &&
+		r.isNilSafeSliceIndex(expr.High, expr.X) &&
+		r.isNilSafeSliceIndex(expr.Max, expr.X)
```
yuxincs: I'm not sure if this makes the code simpler to reason.

the nil safety here depends on `expr.Low`, `expr.High`, and `expr.Max` simultaneously (e.g., `[0:]` is safe, but `[0:1]` is not).

Before, we basically list down 3 cases we are handling and connect them via `||`, which is super clear.

Now after the refactor we are actually passing both the index and expression downwards, where you have to dig deeper to understand.

assertion/function/assertiontree/root_assertion_node.go
```diff
@@ -1215,12 +1215,51 @@ func (r *RootAssertionNode) isType(expr ast.Expr) bool {
...
+	call, ok := ast.Unparen(index).(*ast.CallExpr)
+	if !ok {
+		return false
+	}
+	fun, ok := ast.Unparen(call.Fun).(*ast.Ident)
+	if !ok || !r.isBuiltIn(fun) {
+		return false
+	}
+	switch fun.Name {
+	case "len", "cap":
```
yuxincs: `len(a) == 0` does not imply `cap(a) == 0`, we might want to avoid handling `cap` for now.

assertion/function/assertiontree/root_assertion_node.go
yuxincs: Would you mind elaborating the thought process here? perhaps I missed something

assertion/function/assertiontree/root_assertion_node.go
yuxincs: Also, since these are built-in function, can we refer to https://github.com/uber-go/nilaway/blob/main/util/typeshelper/typeshelper.go and use the builtin helper?

This would help avoid incorrectly matching shadowed `len` user-defined function

assertion/function/assertiontree/root_assertion_node.go
```diff
@@ -1215,12 +1215,51 @@ func (r *RootAssertionNode) isType(expr ast.Expr) bool {
...
+// (nil) and a compile-time zero are always safe, as is an index that provably evaluates to zero when
+// `slice` is nil (see isZeroOnNilIndex).
+func (r *RootAssertionNode) isNilSafeSliceIndex(index, slice ast.Expr) bool {
+	return index == nil || r.Pass().IsZero(index) || r.isZeroOnNilIndex(index, slice)
+}
+
+// isZeroOnNilIndex returns true if `index` provably evaluates to zero when `slice` is nil. Since a
+// nil slice has len == cap == 0, the recognized patterns are `len(slice)` and `cap(slice)`, and
+// `min(..., x, ...)` where any argument `x` is zero on nil.
+func (r *RootAssertionNode) isZeroOnNilIndex(index, slice ast.Expr) bool {
```
yuxincs: should this actually be `isZeroOnNilSlice`?

assertion/function/assertiontree/root_assertion_node.go
karthikbhata97: Yes the cap one is tricky as cap(a) is provably zero only on nil slices.
So I made this function to be `make sure the index is safe on a nil slice` rather than `index is safe on any slice` which works for this usecase. But since this callpath is specific for slicing of nil slices, i thought its reasonable.

But let me know if i should make it more general length bounds check by removing the cap.

assertion/function/assertiontree/root_assertion_node.go
```diff
@@ -1214,13 +1215,60 @@ func (r *RootAssertionNode) isType(expr ast.Expr) bool {
...
+	}
+	fun, ok := ast.Unparen(call.Fun).(*ast.Ident)
+	if !ok {
+		return false
+	}
+	switch r.ObjectOf(fun) {
+	case typeshelper.BuiltinLen:
+		// `len(slice)`, where the argument is the same slice expression being sliced, is zero for a
+		// nil slice.
+		return len(call.Args) == 1 && r.eqStable(call.Args[0], slice)
```
yuxincs, line 1254: I feel `r.eqStable` is a bit of overkill for this purpose, we can simply check if both `call.Args[0]` and `slice` points to the same object (via `r.ObjectOf`)

testdata/src/go.uber.org/slices/slices.go
```diff
@@ -481,6 +481,68 @@ func testSlicingDoesNotCreateConsumersForNilableSlice() []int {
...
+		b = nilA[:min(len(other), 100)] //want "sliced into"
+	case 3:
+		b = nilA[:cap(other)] //want "sliced into"
+	case 4:
+		// max(...) requires every argument to be length-bounded; n and a literal 0 are not, so even
+		// though max(len(x), 0) is genuinely zero on nil, we conservatively create a consumer.
+		b = nilA[:max(len(nilA), 100)] //want "sliced into"
+		b = nilA[:max(len(nilA), 0)]   //want "sliced into"
+	case 5:
+		// cap(x) is no longer treated as length-bounded.
```
yuxincs: why "no longer"?

assertion/function/assertiontree/root_assertion_node.go
```diff
@@ -1214,13 +1215,60 @@ func (r *RootAssertionNode) isType(expr ast.Expr) bool {
...
+			}
+		}
+	case typeshelper.BuiltinMax:
+		// `max(...)` is length-bounded only if all of its arguments are.
+		for _, arg := range call.Args {
+			if !r.isLengthBoundedSliceIndex(arg, slice) {
+				return false
+			}
+		}
+		return len(call.Args) > 0
```
yuxincs, line 1269: Would `max()` be valid? if so can we add that to the test suite?

assertion/function/assertiontree/root_assertion_node.go
```diff
@@ -1214,13 +1215,60 @@ func (r *RootAssertionNode) isType(expr ast.Expr) bool {
...
+// where at least one argument is length-bounded, and `max(...)` where every argument is length-bounded.
+func (r *RootAssertionNode) isLengthBoundedSliceIndex(index, slice ast.Expr) bool {
+	call, ok := ast.Unparen(index).(*ast.CallExpr)
+	if !ok {
+		return false
+	}
+	fun, ok := ast.Unparen(call.Fun).(*ast.Ident)
+	if !ok {
+		return false
+	}
```
yuxincs, line 1249: Can we add test code for paren cases (i.e., to test that we should be able to handle parenthese)?

yuxincs (approved):

assertion/function/assertiontree/root_assertion_node.go
karthikbhata97: Fixed

PR #418, opened 2026-06-11
Model well-known stdlib globals (os.Stdout/Stderr/Stdin, os.Args) as non-nil

yuxincs:
NilAway infers `os.Stdout`, `os.Stderr`, `os.Stdin`, and `os.Args` as potentially nil, because their initializers technically can return nil (`os.Std*` come from `NewFile`, which returns nil for a negative fd; `os.Args` is populated by the runtime). In practice all four are always non-nil, so every use — `io.Copy(os.Stdout, r)`, `os.Stderr.Write(...)`, `os.Args[1]`, etc. -- produces a false positive. On the Go standard library alone this accounts for ~184 of the unique diagnostics.
 
This PR teaches NilAway to treat reads of these globals as non-nil, via the existing `hook` framework.

We also did a bit refactor on the hook framework to allow accepting variables instead of just function and methods.

### DIFF
--- a/hook/assume_return.go
+++ b/hook/assume_return.go
@@ -38,7 +38,7 @@ func AssumeReturn(pass *analysishelper.EnhancedPass, call *ast.CallExpr) *annota
 
 func matchTrustedFuncs(pass *analysishelper.EnhancedPass, call *ast.CallExpr) *annotation.ProduceTrigger {
 	for sig, act := range _assumeReturns {
-		if sig.match(pass, call) {
+		if sig.matchCall(pass, call) {
 			return act(call)
 		}
 	}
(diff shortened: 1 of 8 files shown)


hook/assume_return.go
```diff
@@ -130,38 +130,38 @@ func isErrorWrapperFunc(pass *analysishelper.EnhancedPass, call *ast.CallExpr) b
 
 type assumeReturnAction func(call *ast.CallExpr) *annotation.ProduceTrigger
 
-var _assumeReturns = map[trustedFuncSig]assumeReturnAction{
+var _assumeReturns = map[trustedSig]assumeReturnAction{
```
yuxincs, line 133: The rest of the changes here are just refactors to allow taking variables in our trusted framework

sonalmahajan15 (approved):

Issue #94, opened 2023-11-06
False positive in checking nilabilty of slice using `len(s)` assigned to a variable


sonalmahajan15:
The following code reports a false positive.
```
func test(mp map[string][]*int) {
	sl := mp["foo"]
	slLen := len(sl)
	if slLen == 0 {
		return
	}
	_ = sl[0] // FP: error: deep read from parameter `mp` lacking guarding; sliced
}
```

However, using `len` directly in the condition works as expected, reporting no error.
```
func test(mp map[string][]*int) {
	sl := mp["foo"]
	if len(sl) == 0 {
		return
	}
	_ = sl[0]
}
```

sonalmahajan15:
@abhinav, thanks for reporting this. :) We'll try and add support for this soon.

lazaroclapp:
> 
> However, if `len(slice) - 1 >= 0`, then `len(slice) > 0` so it should infer that the slice is non-nil.
> 

Just to set expectations a bit here, while we are happy to build heuristics for common cases, in the limit NilAway shouldn't be expected to reason about arbitrarily complex integer arithmetic. This is a type system for nilability and that sort of thing is the domain of SMT solvers. 

I do think `x-1 >= 0 => x > 0` can be special cased, though! 😅 

I just wouldn't expect say, this, to ever work:

```
j := len(b.bs)
i := (j - 1) % 2
if i >= 0 {
  b.bs[i] = ...
}
```

Even though, of course, it's equally safe (len(nil) = 0 => i =  (0-1) % 2 = -1).

In theory a tool can be built that reasons about whether values can contain nil and the possible value ranges of every integer variable in the program, but the later is a harder problem than the former.

abhinav:
Thanks, @lazaroclapp, that's completely reasonable.

I'm aware that not requiring code annotations was one of the goals of nilaway, but just thinking out loud:
Would there be room in the system to annotate code with comments for these corner cases?

For example, for the Zap function above:

```go
func (b *Buffer) TrimNewline() {
	if i := len(b.bs) - 1; i >= 0 {
		if b.bs[i] == '\n' {
			b.bs = b.bs[:i]
		}
	}
}
```

If authors could do something like this, informing nilaway that that condition implies something about the result:

```go
if i := len(b.bs) - 1; i >= 0 {
	// @implies:not-nil(b.bs) // i > 0 guarantees b.bs is non-nil
	if b.bs[i] == '\n' {
		// ...
	}
}
```

[...]

PR #419, opened 2026-06-11
Track `len(s) - positive >= 0` / `len(s) + negative >= 0` as a non-nil check in AddNilCheck

karthikbhata97:
Adds a sound heuristic to `AddNilCheck` so that a conditional that guarantees a slice/map has at least one element marks it as non-nil in the guarded branch. This addresses the false positive called out in #94.

The supported forms are any expression that reduces to `len(s) - k >= 0` with integer `k >= 1`:
- `len(s) - c >= 0` where `c` is a **positive** constant
- `len(s) + c >= 0` where `c` is a **negative** constant

In both cases `len(s) - k >= 0` implies `len(s) >= k >= 1`, so `s` cannot be nil in that branch.

## Details
- New checker (op `token.GEQ`) in `AddNilCheck` backed by the `lenMinusPositiveArg` helper. The existing driver loop derives the converse/inverse/contrapositive automatically, so the following forms are covered for free (shown for `len(a) - 1`):
  - `0 <= len(a) +- c`
  - `len(a) +- c < 0` (non-nil in the false branch)
  - `0 > len(a) +- c` (non-nil in the false branch)

## Soundness
The helper deliberately keeps the match tight so we never conclude non-nil unsoundly:
- The `len(...)` term must be on the **left**, so `c - len(s) >= 0` (which only bounds the length from above) is rejected.
- The offset must be a **constant**, so `len(s) - 1 + b >= 0` is rejected.
- `k >= 1` is required, so the always-true / no-information cases `len(s) - 0 >= 0` and `len(s) + 1 >= 0` are rejected.

## Scope
Limited to the constant-offset arithmetic form above. The broader case from #94 with the length captured into a variable first (`n := len(s); if n == 0 { return }`) is out of scope.

## Testing
- Added cases in `testdata/src/go.uber.org/slices/slices.go` covering the supported forms (`len(a) - 1`, `len(a) - 5`, `len(a) + (-1)`, `len(a) + (-3)`, and their automatic flips, all expecting no error) plus negative cases that must still report (`len(a) - 0 >= 0`, `len(a) + 1 >= 0`, `5 - len(a) >= 0`).

### DIFF
--- a/assertion/function/assertiontree/util.go
+++ b/assertion/function/assertiontree/util.go
@@ -279,6 +279,24 @@ func AddNilCheck(pass *analysishelper.EnhancedPass, expr ast.Expr) (trueCheck, f
 				return noop, noop, true
 			},
 		},
+		{
+			// `len(a) - 1 >= 0`
+			// Unlike the looser `len(a) ... >= [positive-int]` heuristic above, this is sound:
+			// `len(a) - 1 >= 0` implies `len(a) >= 1`, so `a` is non-nil. We require the exact
+			// `len(a) - 1` structure (rather than the loose `extractLenArgs` matching) so that we
+			// do not unsoundly conclude anything from, e.g., `len(a) - 1 + b >= 0`.
+			// Automatic cases:
+			//   - `0 <= len(a) - 1`
+			//   - `len(a) - 1 < 0`
+			//   - `0 > len(a) - 1`
+			op: token.GEQ,
+			matcher: func(x, y ast.Expr) (RootFunc, RootFunc, bool) {
+				if arg, ok := lenMinusOneArg(pass, x); ok && pass.IsZero(y) {
+					return produceNegativeNilChecks(arg), noop, false
+				}
+				return noop, noop, true
+			},
+		},
 	}
 
 	// Apply the checkers.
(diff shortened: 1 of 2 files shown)


assertion/function/assertiontree/util.go
```diff
@@ -279,6 +279,21 @@ func AddNilCheck(pass *analysishelper.EnhancedPass, expr ast.Expr) (trueCheck, f
...
+			//   - `len(a) - 1 < 0`
+			//   - `0 > len(a) - 1`
+			op: token.GEQ,
+			matcher: func(x, y ast.Expr) (RootFunc, RootFunc, bool) {
+				if arg, ok := lenMinusOneArg(pass, x); ok && pass.IsZero(y) {
+					return produceNegativeNilChecks(arg), noop, false
+				}
+				return noop, noop, true
+			},
+		},
```
yuxincs: Hmmmm, it seems that from the case above we should've had handling for this, can we figure out why that's not happening instead of introducing another case?

karthikbhata97: There are checker that perform `> 0` and `>= positive` and none do `>= 0`, hence it was not catching this case.
Also, those checkers are unsound, whereas this specific one is sound.

Do you suggest keeping it separate as this is a sound check, or transforming the `>= positive` to `>= [0 or positive]`?

assertion/function/assertiontree/util.go
```diff
@@ -279,6 +279,21 @@ func AddNilCheck(pass *analysishelper.EnhancedPass, expr ast.Expr) (trueCheck, f
 				return noop, noop, true
 			},
 		},
+		{
```
karthikbhata97, line 282: @yuxincs these cases, i referred in the other comment

assertion/function/assertiontree/util.go
yuxincs: Ah, I see!

Then let's keep this separate!

Should we slightly generalize the checker to handle `len(a) - [positive int] >= 0`, and perhaps `len(a) + [negative int] >= 0`?

karthikbhata97: The unsound checkers are actually super general, all the check is if `len` is called on one side and anything accompanying them (uses ast.Inspect recursively) and `> 0 or positive` and `>= positive`. (even though the comments there say `len(a) - 1 + b` on lhs)

So if we are to generalize this, it probably is just merging those two into `>= 0 or positive`. But maybe not worth introducing more unsound-ness unless there are reports of frequent such false positives?

yuxincs (approved):

PR #420, opened 2026-06-12
Model testify, gotest.tools, and goconvey assertions as trusted functions

TvdW:
This PR adds three more test libraries to the hooks, in an attempt to further reduce false positives around test suites.

No diff on stdlib expected because it doesn't use any of these 😉

### DIFF
--- a/hook/split_blocks_on.go
+++ b/hook/split_blocks_on.go
@@ -22,6 +22,7 @@ import (
 	"regexp"
 
 	"go.uber.org/nilaway/util/analysishelper"
+	"go.uber.org/nilaway/util/asthelper"
 	"go.uber.org/nilaway/util/typeshelper"
 )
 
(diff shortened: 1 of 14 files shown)


hook/split_blocks_on.go
```diff
@@ -80,6 +81,93 @@ var negatedSelfExpr splitBlockOnAction = func(_ *analysishelper.EnhancedPass, ca
...
+
+// isBoolExpr reports whether the expression is statically of boolean type.
+func isBoolExpr(pass *analysishelper.EnhancedPass, expr ast.Expr) bool {
+	t := pass.TypesInfo.TypeOf(expr)
+	if t == nil {
+		return false
+	}
+	basic, ok := t.Underlying().(*types.Basic)
+	return ok && basic.Kind() == types.Bool
+}
```
yuxincs: Attach this to `analysishelper.EnhancedPass` as a method?

TvdW: Oh that makes sense. Done!

hook/split_blocks_on.go
```diff
@@ -80,6 +81,83 @@ var negatedSelfExpr splitBlockOnAction = func(_ *analysishelper.EnhancedPass, ca
...
+	}
+	var ident *ast.Ident
+	switch assert := call.Args[argIndex+1].(type) {
+	case *ast.Ident:
+		ident = assert
+	case *ast.SelectorExpr:
+		ident = assert.Sel
+	default:
+		return nil
+	}
```
yuxincs, line 135: This seems to be used quite often throughout nilaway, we probably want to extract this to a helper function or whatnot. but we can leave that for later :)

yuxincs (approved):

PR #421, opened 2026-06-12
Re-insert type switch case conditions into the CFG

TvdW:
The CFG builder (golang.org/x/tools/go/cfg) represents each case type of a type switch as an empty two-successor branching block, with no expression for the implicit per-case type assertion. NilAway re-inserts the lost conditions for value switches (markSwitchStatements) but had nothing equivalent for type switches, so `case nil` arms were not recognized as nil guards:

    ptr, err := f()
    switch err.(type) {
    case nil:
        _ = *ptr // FP: err is provably nil here, so ptr is non-nil
    }

Real-world hits include gotest.tools/assert.NilError (reported inside gotest.tools, where it cannot be suppressed) and
protoreflect.ValueOf(nil).

The new markTypeSwitchStatements pass synthesizes `operand == nil` / `operand != nil` conditions for nil and interface-typed case arms, and replicates the binding assignment (`x := v.(type)`) into each case body so that assertions on `x` transfer to `v` while per-branch knowledge is available; see its doc comment for the full semantics and soundness reasoning. Arms are classified via the type checker (EnhancedPass.IsNil now consults TypesInfo before the name-based fallback needed for synthesized identifiers), which also hardens source-level nil checks against a user-shadowed `nil`.

PR #366 previously attempted this by matching `(x.(type)) == nil`, but no such conditions are ever synthesized, so its testdata passed vacuously. That dead special case is removed and the testdata made genuinely fallible, then extended with soundness tests: default arms, multi-type clauses, concrete and type-parameter arms, parenthesized `case (nil)`, and a shadowed `nil` type (package shadownil).

### DIFF
--- a/assertion/function/preprocess/cfg.go
+++ b/assertion/function/preprocess/cfg.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"go/ast"
 	"go/token"
+	"go/types"
 
 	"go.uber.org/nilaway/hook"
 	"go.uber.org/nilaway/util/asthelper"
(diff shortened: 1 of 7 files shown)


assertion/function/preprocess/cfg.go
```diff
@@ -87,13 +88,15 @@ func (p *Preprocessor) CFG(graph *cfg.CFG, funcDecl *ast.FuncDecl) *cfg.CFG {
...
+	// Next, we need to re-insert information that is lost during CFG build for *ast.RangeStmt,
+	// *ast.SwitchStmt, and *ast.TypeSwitchStmt by iterating through all blocks. This requires
+	// knowing the links between the nodes contained within a block to their parents
+	// (*ast.RangeStmt, *ast.SwitchStmt, or *ast.TypeSwitchStmt nodes).
 	// So, here establish the link and then do the work.
-	rangeChildren, switchChildren := collectChildren(funcDecl)
+	rangeChildren, switchChildren, typeSwitchChildren := collectChildren(funcDecl)
 	markRangeStatements(graph, rangeChildren)
 	markSwitchStatements(graph, switchChildren)
+	p.markTypeSwitchStatements(graph, typeSwitchChildren)
```
yuxincs, line 99: nit: does this have to be attached to `p` as a method? or can they simply be functions like the other two?

yuxincs (approved):

yuxincs:
@TvdW I just have a nit but otherwise LGTM, thanks!

assertion/function/preprocess/cfg.go
TvdW: Unfortunately yes: the code accesses `p.pass.IsNil` and `p.pass.TypesInfo` later on, to identify the type of each case arm. I could pass `p.pass` down the stack- but I'm not sure that's really better

yuxincs: Ah, OK, then I would actually prefer to attach them as methods.

Makes sense to me!

PR #422, opened 2026-06-15
typeshelper: follow type aliases, generics, to resolve underlying types

TvdW:
While doing some prework for my next commit I realized that these don't resolve correctly, hiding some types from being checked for nilness.

### DIFF
--- a/assertion/function/assertiontree/backprop.go
+++ b/assertion/function/assertiontree/backprop.go
@@ -491,13 +491,13 @@ func backpropAcrossRange(rootNode *RootAssertionNode, lhs []ast.Expr, rhs ast.Ex
 	case 1:
 		// If we are ranging over a map slice or string with only a single lhs operand, then that
 		// operand will be int-valued.
-		if typeshelper.IsDeeplyMap(rhsType) || typeshelper.IsDeeplySlice(rhsType) || typeshelper.IsDeeplyArrayOrArrayPtr(rhsType) || typeIsString(rhsType) {
+		if typeshelper.IsDeeplyType[*types.Map](rhsType) || typeshelper.IsDeeplyType[*types.Slice](rhsType) || typeshelper.IsDeeplyArrayOrArrayPtr(rhsType) || typeIsString(rhsType) {
 			produceAsIndex(0)
 			return nil
 		}
 		// Iterating over a channel with only a single lhs operand will still result in deeply
 		// produced lhs values.
-		if typeshelper.IsDeeplyChan(rhsType) {
+		if typeshelper.IsDeeplyType[*types.Chan](rhsType) {
 			produceAsDeepRHS(0)
 			return nil
 		}
--- a/assertion/function/assertiontree/backprop.go
+++ b/assertion/function/assertiontree/backprop.go
@@ -491,7 +491,7 @@ func backpropAcrossRange(rootNode *RootAssertionNode, lhs []ast.Expr, rhs ast.Ex
 	case 1:
 		// If we are ranging over a map slice or string with only a single lhs operand, then that
 		// operand will be int-valued.
-		if typeshelper.IsDeeplyMap(rhsType) || typeshelper.IsDeeplySlice(rhsType) || typeshelper.IsDeeplyArray(rhsType) || typeIsString(rhsType) {
+		if typeshelper.IsDeeplyMap(rhsType) || typeshelper.IsDeeplySlice(rhsType) || typeshelper.IsDeeplyArrayOrArrayPtr(rhsType) || typeIsString(rhsType) {
 			produceAsIndex(0)
 			return nil
 		}
--- a/assertion/function/assertiontree/parse_expr_producer.go
+++ b/assertion/function/assertiontree/parse_expr_producer.go
@@ -479,7 +479,7 @@ func (r *RootAssertionNode) ParseExprAsProducer(expr ast.Expr, doNotTrack bool) 
 		// storage. For example, `var a [4]int; _ = a[:0]` is a nonnil (empty) slice. This holds
 		// regardless of the indices, so we must check it before the `b[_:0:_]` case below (which
 		// would otherwise wrongly treat `a[:0]` as a nilable empty slice).
-		if typeshelper.IsDeeplyArray(r.Pass().TypesInfo.Types[expr.X].Type) {
+		if typeshelper.IsDeeplyArrayOrArrayPtr(r.Pass().TypesInfo.TypeOf(expr.X)) {
 			// Returning nil to indicate the slice expression results in a nonnil slice.
 			return nil, nil
 		}
--- a/util/typeshelper/typeshelper.go
+++ b/util/typeshelper/typeshelper.go
@@ -75,99 +75,137 @@ func IsSlice(t types.Type) bool {
 	}
 }
 
-// IsDeeplyArray returns true if `t` is of array type, including
-// transitively through Named types
+// IsDeeplyArray returns true if `t` is of array type, including transitively through named
+// types and aliases, as well as type parameters whose type sets contain only array types.
 func IsDeeplyArray(t types.Type) bool {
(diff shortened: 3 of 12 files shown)


assertion/function/assertiontree/backprop.go
```diff
@@ -491,7 +491,7 @@ func backpropAcrossRange(rootNode *RootAssertionNode, lhs []ast.Expr, rhs ast.Ex
 	case 1:
 		// If we are ranging over a map slice or string with only a single lhs operand, then that
 		// operand will be int-valued.
-		if typeshelper.IsDeeplyMap(rhsType) || typeshelper.IsDeeplySlice(rhsType) || typeshelper.IsDeeplyArray(rhsType) || typeIsString(rhsType) {
+		if typeshelper.IsDeeplyMap(rhsType) || typeshelper.IsDeeplySlice(rhsType) || typeshelper.IsDeeplyArrayOrArrayPtr(rhsType) || typeIsString(rhsType) {
```
yuxincs: Can we use the UnwrapPtr helper function and just call `typeshelper.IsDeeplyArray(typeshelper.UnwrapPtr(rhsType))` such that we don't have to introduce a new API in `typeshelper`?

Or did I miss anything here?

assertion/function/assertiontree/parse_expr_producer.go
```diff
@@ -479,7 +479,7 @@ func (r *RootAssertionNode) ParseExprAsProducer(expr ast.Expr, doNotTrack bool)
 		// storage. For example, `var a [4]int; _ = a[:0]` is a nonnil (empty) slice. This holds
 		// regardless of the indices, so we must check it before the `b[_:0:_]` case below (which
 		// would otherwise wrongly treat `a[:0]` as a nilable empty slice).
-		if typeshelper.IsDeeplyArray(r.Pass().TypesInfo.Types[expr.X].Type) {
+		if typeshelper.IsDeeplyArrayOrArrayPtr(r.Pass().TypesInfo.TypeOf(expr.X)) {
```
yuxincs, line 482: 👍

util/typeshelper/typeshelper.go
```diff
@@ -75,99 +75,137 @@ func IsSlice(t types.Type) bool {
...
 func IsDeeplyArray(t types.Type) bool {
-	switch tt := UnwrapPtr(t).(type) {
-	case *types.Array:
-		return true
-	case *types.Named:
-		return IsDeeplyArray(tt.Underlying())
-	}
-	return false
+	return underlyingIs[*types.Array](t)
 }
```
yuxincs: Now that we have `underlyingIs`, can we remove these APIs from the good old times and just use the generic version instead?

Of course, we should do that in a separate PR to avoid bloating this one :)

util/typeshelper/typeshelper.go
```diff
@@ -75,99 +75,137 @@ func IsSlice(t types.Type) bool {
...
+	})
+}
+
+// underlyingAlwaysSatisfies reports whether the underlying type of `t` satisfies pred. Named
+// types and aliases are resolved via Underlying(). For type parameters, the underlying type of
+// every term in the constraint's type set must satisfy pred. Since the elements of a constraint
+// interface intersect, this is conservative: the actual type set can only be smaller than the
+// enumerated terms, and type sets we cannot enumerate (such as method-only constraints) yield
+// false.
+func underlyingAlwaysSatisfies(t types.Type, pred func(types.Type) bool) bool {
```
yuxincs, line 108: It seems that we only ever have a single `pred`, do we foresee that we will add more predicates in the future? if not I'd actually prefer we inline it

util/typeshelper/typeshelper.go
```diff
@@ -75,99 +75,137 @@ func IsSlice(t types.Type) bool {
...
+			if emb, isIface := e.Underlying().(*types.Interface); isIface {
+				terms = append(terms, constraintTerms(emb)...)
+			} else {
+				terms = append(terms, e)
+			}
+		}
 	}
-	return false
+	return terms
+}
```
yuxincs: Do we know how much of a performance penalty this would bring?

We'll be calling this function over and over again (and perhaps for the same type), if this makes it slow we should probably at least add a cache.

Or, we can also move this part out to a separate PR and I'm happy to merge the rest first while we discuss on this.

yuxincs:
@TvdW This is an awesome PR! I left some comments, let me know if they don't make sense :)

assertion/function/assertiontree/backprop.go
TvdW: Not quite: UnwrapPtr only takes care of the pointer, but no type aliases, named types, parameters, etc

util/typeshelper/typeshelper.go
TvdW: Happy to submit a follow-up PR! `func IsType[T](t types.Type) bool`

util/typeshelper/typeshelper.go
TvdW: fwiw I can remove this function entirely if you're ok with me adding a new dependency on https://pkg.go.dev/golang.org/x/exp/typeparams#NormalTerms (which does a slightly better job as well, I just wasn't sure if you'd be ok with a new dependency.)

perf-wise, this code only runs on structs with type parameters. it doesn't seem to hurt the golden test- but I admittedly haven't measured the performance impact on code that heavily uses generics (like Rust-style code)

util/typeshelper/typeshelper.go
TvdW: hm, not trivially. this function has two callers with different predicates (IsDeeplyArrayOrArrayPtr and underlyingIs), and the predicate is used to check type parameters so it needs to loop. I could create a new function `getUnderlyingTypes` but it would need to return a freshly allocated slice which seems worse?

util/typeshelper/typeshelper.go
TvdW: I ran a benchmark using the uncached version vs `map[types.Type]...` vs `sync.Map`. the uncached underlyingIs outperformed its cached version (respectively 5ns/op, 9ns/op, 7ns/op) for non-parameterized types. With parameters this quickly jumped to 100ns and even upwards of 500ns for more complex parameters. Still: are those common enough to warrant it?

assertion/function/assertiontree/backprop.go
yuxincs: I have a feeling that we can make `UnwrapPtr` take care of type aliases, named types, parameters and it would just work.

But let's leave that to a separate journey, the blast radius is large and we may hit corner cases.

Agreed here!

util/typeshelper/typeshelper.go
yuxincs: Yes exactly! Please do!

util/typeshelper/typeshelper.go
yuxincs: Huh, OK then let's keep this as-is for now :)

util/typeshelper/typeshelper.go
yuxincs: Yes, having dependencies from x/exp is totally OK (I really wish `NormalTerms` can be included in stdlib, maybe in the future -- then we can drop that dependency)

yuxincs: > Still: are those common enough to warrant it?

Yeah I agree that generics in Go aren't that popular (yet?) and this should be OK. Let's swap to `NormalTerms` and merge. 

I can run this on our internal benchmarks too, if needed we can always revert this part

yuxincs:
@TvdW Replied to all comments, please ping me once this PR is updated and I'm happy to merge :)

assertion/function/assertiontree/backprop.go
TvdW: That's #431 😁

TvdW:
> @TvdW Replied to all comments, please ping me once this PR is updated and I'm happy to merge :)

@yuxincs Done!

yuxincs (approved):

PR #423, opened 2026-06-15
Model implicit pointer-to-array dereferences as nil consumers

TvdW:
Slicing, indexing, and 2-variable ranging over a *[N]T implicitly dereference the pointer (e.g. p[low:high] is shorthand for (*p)[low:high]), so they panic on a nil pointer. Notably this holds even for slice forms like p[:0] and p[:] that are safe on a nil slice. len/cap and range loops with at most one (possibly blank) iteration variable never read the elements and stay nil-safe.

Add PtrLoad consumers for these cases in consumeIndexExpr, the SliceExpr case of AddComputation (bypassing the slice-only safe-slicing carve-out), and case 2 of backpropAcrossRange.


TvdW:
In my defense, the golden test 📈 means the code works 😉

testdata/src/go.uber.org/slices/inference/arrayptr.go
```diff
@@ -0,0 +1,107 @@
...
+func nilArrayPtr() *[4]int {
+	if arrayPtrDummy {
+		return nil
+	}
+	return &[4]int{}
+}
+
+func testArrayPtrSliceSafeForm() []int {
+	p := nilArrayPtr()
+	return p[:0] //want "dereferenced"
```
yuxincs, line 35: I wonder if we should use a more specialized consumer such that we can use a more friendly message.

I.e., it slipped through our minds that this is actually panic-able. We probably want to say a little more here to the end users such that they understand.

What do you think?

yuxincs (commented):
This is awesome! Thanks! I left a comment as a discussion point.

testdata/src/go.uber.org/slices/inference/arrayptr.go
TvdW: Hm, interesting question. I'm inclined to say that while the dereference is implicit (there's no `*`, but an index expression or range), the type in the user's code is explicitly a pointer (save for type aliasing etc), so it shouldn't cause too much confusion.

yuxincs: Thinking a bit more I think I tend to agree :) 

Thanks, merging!

yuxincs (approved):

PR #424, opened 2026-06-15
Fix contract inference using the receiver instead of the declared parameter

TvdW:
In SSA, fn.Params[0] is the receiver for methods, while contract inference gates on (and call-site application uses) the declared parameter. As a result, fluent methods returning their receiver (e.g., `func (z *Int) Neg(x *Int) *Int { ...; return z }`) incorrectly matched the param == ret rule via the receiver and were given contract(nonnil -> nonnil) about `x`. Skip the receiver when selecting the parameter to verify the contract against.


yuxincs (approved):
Nice catch! Thanks!

PR #425, opened 2026-06-16
Improve MergeFullTriggers to avoid new allocation

karthikbhata97:
## Summary
~10% faster wall-clock and −14% total allocations (−12.9 GB) on full stdlib, output unchanged.

## Method
- Workload — the project's own golden-test stdlib command:
  `nilaway -include-errors-in-files / -json -pretty-print=false -group-error-messages=true std`
- Hardware: Apple Silicon, 16 cores, `GOMAXPROCS=16`. 1 warm-up + 8 measured iterations per variant,
  **interleaved** (base, fix, base, fix …) to cancel thermal/background drift.

## Correctness
Output is **byte-for-byte identical** between base and fix (2,091,004 bytes of JSON diagnostics).
A perf-only change, no behavior change.

## Wall-clock & CPU (8 interleaved iterations)

| metric | base median | fix median | Δ |
|--------|------------:|-----------:|----:|
| real (wall) | 15.11 s | 13.63 s | **−9.8%** |
| user (CPU)  | 182.15 s | 162.88 s | **−10.6%** |
| sys (CPU)   | 11.50 s | 10.40 s | −9.6% |

Distributions do not overlap (base min 14.57 s > fix max 13.87 s) and the fix is also less noisy
(real σ: 0.44 s → 0.13 s). ~10% wall-clock speedup on full stdlib.

## Allocation / GC churn (heap profile, `alloc_space`)

| | total allocated over run |
|--|--:|
| base | 89.76 GB |
| fix  | 76.87 GB |
| **Δ** | **−12.89 GB (−14.4%)** |

`MergeFullTriggers` flat allocations (per `pprof -list`):
- base: **12.80 GB** (12.36 GB on the old `out = append(out, l)` full-copy line + 0.44 GB)
- fix:  **0.61 GB** (in-place `left = append(left, r)`)

The ~12.2 GB reduction in this single function accounts for essentially the entire 12.9 GB total
drop — direct confirmation that the per-call copy of `left` was the dominant allocation source.

### DIFF
--- a/annotation/full_trigger.go
+++ b/annotation/full_trigger.go
@@ -168,8 +168,8 @@ func FullTriggerSlicesEq(left, right []FullTrigger) bool {
 // Consumer.GuardMatched into a single trigger with Consume.GuardMatched = false. In all other cases - such as
 // checking fixed point in propagation, the function FullTriggersEq
 // that does observe GuardMatched should be used instead of this function.
+// This function modifies the `left` slice, by appending the non-duplicate right triggers into it.
 func MergeFullTriggers(left []FullTrigger, right ...FullTrigger) []FullTrigger {
-	// We append non-duplicate right triggers in place into `left` slice.
 	// `left` is dedup-maintained by construction, so we compare each right trigger only against the
 	// original `left` prefix (origLen) and stop at the first match.
 	origLen := len(left)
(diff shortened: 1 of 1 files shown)


annotation/full_trigger.go
```diff
@@ -169,40 +169,30 @@ func FullTriggerSlicesEq(left, right []FullTrigger) bool {
 // checking fixed point in propagation, the function FullTriggersEq
 // that does observe GuardMatched should be used instead of this function.
 func MergeFullTriggers(left []FullTrigger, right ...FullTrigger) []FullTrigger {
```
karthikbhata97, line 172: @yuxincs I did not change the function into method as go does not allow defining a method for a type (slice here) defined outside the package
Also I kept the same signature, as it mimics golang's `append` style, updating and returning 

Let me know if you want me to rename it differently.

yuxincs: Ah, that makes sense. Can we at least move the comments to the function docstring stating that the input slice will be modified?

yuxincs (approved):

PR #430, opened 2026-06-29
Fix false positive for errors.As guard in right operand of &&

TvdW:
When errors.As(err, &target) appears as the left operand of && in a non-conditional context (e.g. assignment or return statement), go/cfg represents the whole statement as a single flat CFG node, so the existing replaceConditional preprocessing never fires and the right operand has no nil guard for target.

Fix by extending AddNilCheck to handle *ast.CallExpr: if ReplaceConditional recognises the call (e.g. errors.As → "call && target != nil"), we extract and return the nil check from the right operand of that replacement. This lets the existing LAND handling in AddComputation propagate the guard to field/selector accesses in the right operand without any CFG restructuring.

Adds test cases for the method-call and field-access variants of the pattern.

### DIFF
--- a/testdata/src/go.uber.org/trustedfunc/trustedfuncs.go
+++ b/testdata/src/go.uber.org/trustedfunc/trustedfuncs.go
@@ -984,6 +984,12 @@ func errorsAs(err error, num string, dummy bool) {
 		if errors.As(err, &exitErr) && dummy {
 			print(*exitErr)
 		}
+	case "method call on target in short-circuit AND (right operand)":
+		var exitErr *exec.ExitError
+		_ = errors.As(err, &exitErr) && exitErr.Exited()
+	case "field access on target in short-circuit AND (right operand)":
+		var myErr *exec.Error
+		_ = errors.As(err, &myErr) && myErr.Name == ""
 	case "errors.As with other conditionals connected by OR":
 		var exitErr *exec.ExitError
 		if errors.As(err, &exitErr) || dummy {
--- a/assertion/function/assertiontree/util.go
+++ b/assertion/function/assertiontree/util.go
@@ -22,6 +22,7 @@ import (
 	"slices"
 
 	"go.uber.org/nilaway/annotation"
+	"go.uber.org/nilaway/hook"
 	"go.uber.org/nilaway/util/analysishelper"
 	"go.uber.org/nilaway/util/tokenhelper"
 	"go.uber.org/nilaway/util/typeshelper"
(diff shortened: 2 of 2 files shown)


TvdW:
@yuxincs Sorry for the wave of PRs, but I found another one 😄 I think that's number six?

For context, `rules_go` (Bazel) recently fixed an issue that stopped it from analyzing external packages, and it triggered a wave of hundreds of new nilaway findings across our codebase (including many true positives!). Rather than silencing the false positives, I figured I'd ask Claude to have a look to see if we can solve those for everyone.

yuxincs:
@TvdW No that's great! We love contributions :)

I'll be a bit slow these days but will get to review & merge them.

yuxincs:
> recently fixed an issue that stopped it from analyzing external packages

Would you have some pointers to the PR? just out of curiosity. But this does mean that we should assume external dependencies are not analyzed by default (well, we always have such an assumption, but this serves as an even stronger evidence) and add more modeling for 3rd party popular packages.

TvdW:
> > recently fixed an issue that stopped it from analyzing external packages
> 
> Would you have some pointers to the PR? just out of curiosity. But this does mean that we should assume external dependencies are not analyzed by default (well, we always have such an assumption, but this serves as an even stronger evidence) and add more modeling for 3rd party popular packages.

This is that PR: https://github.com/bazel-contrib/rules_go/pull/4615

testdata/src/go.uber.org/trustedfunc/trustedfuncs.go
```diff
@@ -984,6 +984,12 @@ func errorsAs(err error, num string, dummy bool) {
 		if errors.As(err, &exitErr) && dummy {
 			print(*exitErr)
 		}
+	case "method call on target in short-circuit AND (right operand)":
+		var exitErr *exec.ExitError
+		_ = errors.As(err, &exitErr) && exitErr.Exited()
+	case "field access on target in short-circuit AND (right operand)":
+		var exitErr *exec.ExitError
+		_ = errors.As(err, &exitErr) && len(exitErr.Stderr) == 0
```
yuxincs: We should add a dereference to really trigger the panic, right?

yuxincs: Also, we do canonicalize the CFG such that `if foo && bar` are transformed to `if foo { if bar {}}` to explicitly combat cases like this, I'm wondering what makes NilAway miss this 

https://github.com/uber-go/nilaway/blob/136c083a9b7705454c8454358160bb4ec4f415e7/assertion/function/preprocess/cfg.go#L277-L281

🤔

TvdW: Ha, I can see why this is confusing. The dereference here was accessing the `Stderr` field of `*exec.ExitError`. I've updated this to read a `Name` field of `*exec.Error` instead to remove the confusion.

```
        analysistest.go:654: go.uber.org/trustedfunc/trustedfuncs.go:992:33: unexpected diagnostic: Potential nil panic detected. Observed nil flow from source to dereference point:
            	- trustedfunc/trustedfuncs.go:992:33: unassigned variable `myErr` accessed field `Name`
```

TvdW: canonicalizeConditional doesn't reach this code because these tests happen in an assign/return statement, not `if`

yuxincs: Ahhhhh OK I see now, let me do another pass real quick. Thanks for the clarifications!

yuxincs (approved):

PR #431, opened 2026-07-02
typeshelper: be even more thorough in resolving types

TvdW:


TvdW:
@yuxincs it's a bigger change, and comes with a +5% diff on the golden tests... I've put this PR up for visibility into what a followup to #422 could look like, if you're interested (review individual commits, not the diff as a whole; I can put them up as separate stacked PRs if desired)

PR #434, opened 2026-07-07
assertiontree: apply append deep-assignment handling to named slice types

TvdW:
The x = append(x, ...) handling used the shallow IsSlice check, so parameters of named slice types (or aliases) silently skipped deep-assignment analysis. Use IsDeeplyType[*types.Slice] instead and drop the now-unused IsSlice. For named slice types the diagnostic attaches to the type's own annotation site rather than the parameter.


sonalmahajan15 (approved):
LGTM

PR #436, opened 2026-07-07
feat(struct-init-v2): wire struct field effects analyzer into nilaway

karthikbhata97:
Wire the new structfieldeffects analyzer into NilAway’s function analysis so struct-init-v2 boundary bindings use the precomputed package-level field-read summary.

Changes
- Add structfieldeffects.Analyzer as a dependency of the function analyzer.
- Pass *structfieldeffects.ParamFieldEffects into assertiontree.FunctionContext.
- Add read-path accessors on ParamFieldEffects for parameter and return boundary demand.
- Add struct-init-v2 assertion-tree binding logic for allocation fields, return fields, parameter fields, and call arguments.
- Enable the currently supported structinitv2 test packages in TestStructInitV2.

### DIFF
--- a/assertion/function/assertiontree/backprop.go
+++ b/assertion/function/assertiontree/backprop.go
@@ -33,6 +33,7 @@ import (
 	"go.uber.org/nilaway/util/asthelper"
 	"go.uber.org/nilaway/util/typeshelper"
 	"golang.org/x/tools/go/cfg"
+	"golang.org/x/tools/go/types/typeutil"
 )
 
 // backpropAcrossBlock iterates over all nodes in the CFG block in _reverse_ order, writes logs,
@@ -153,6 +154,10 @@ func backpropAcrossReturn(rootNode *RootAssertionNode, node *ast.ReturnStmt) err
 
 	if rootNode.functionContext.functionConfig.EnableStructInitCheck {
 		rootNode.addConsumptionsForFieldsOfParams()
+	}
+
+	if rootNode.functionContext.functionConfig.EnableStructInitV2 {
+		rootNode.bindReturnFieldsToContext(node)
 	}
 
 	if len(node.Results) == 1 {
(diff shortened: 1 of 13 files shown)


assertion/function/assertiontree/backprop.go
```diff
@@ -806,6 +821,18 @@ func backpropAcrossManyToOneAssignment(rootNode *RootAssertionNode, lhs, rhs []a
 			continue
 		}
 
+		// Bind return fields before LHS production detaches its subtree.
+		if rootNode.functionContext.functionConfig.EnableStructInitV2 {
+			if funcObj := typeutil.StaticCallee(rootNode.Pass().TypesInfo, rhsVal); funcObj != nil {
```
yuxincs, line 826: IIRC `typeutil.StaticCallee` is quite expensive, but we can probably optimize this later

testdata/src/go.uber.org/structinitv2/defaultfield/defaultfield.go
```diff
@@ -55,15 +55,6 @@ func escapeIntoSink() {
...
 
-// A nil field escapes via a callee's side effect and is dereferenced by the caller afterwards.
-func clobber(c *A) { c.aptr = nil }
-
-func derefAfterSideEffect() {
-	b := &A{aptr: &A2{}}
-	clobber(b)
-	print(b.aptr.ptr) //want "field `aptr`"
-}
-
```
yuxincs, line 66: Any reason we are removing the test code here?

yuxincs (approved):

testdata/src/go.uber.org/structinitv2/defaultfield/defaultfield.go
karthikbhata97: Since i split the code into multiple PRs, this specific handling (mutated param effect) is not yet added so this will fail. So i have moved this to limitations for now, and will include this once that change is included.

PR #437, opened 2026-07-10
feat(struct-init-v2): propagate field effects across packages

karthikbhata97:
Extend the structfieldeffects analyzer to export and import parameter field-read summaries across package boundaries, using Go's analysis Facts mechanism, so cross-package calls no longer under-report which fields a callee dereferences.

#### Changes
- Add `ParamFieldReadsPackageFact` (`assertion/function/structfieldeffects/fact.go`), a gob-encodable per-package fact holding each exported function's sorted parameter field-read paths.
- Export the fact from the analyzer for every exported function with a non-empty read set, keyed by objectpath of the function for stable cross-package lookup.
- Import facts for statically-resolvable cross-package callees before running `closeParamFieldSets`, seeding their param reads via the new `seedImportedParamReads`, so forwarding closure accounts for reads that happen in the callee's own package.
- Track callees encountered during `collectParamForwardEdges` so the analyzer only imports facts for packages actually called into.
- Export `indexedFieldPath` as `IndexedFieldPath` so it can be serialized as part of the fact.
- Add `fact_test.go` covering export, import, import-with-real serialization (via `analysistest`/`checker.Analyze`), and gob codec determinism, plus a codec benchmark.
- Add upstream/downstream testdata packages under `factexport/` exercising a cross-package forwarded read.
- Enable `go.uber.org/structinitv2/crosspkg/app` in `TestStructInitV2`.

### DIFF
--- a/assertion/function/structfieldeffects/analyzer.go
+++ b/assertion/function/structfieldeffects/analyzer.go
@@ -18,11 +18,16 @@
 package structfieldeffects
 
 import (
+	"cmp"
+	"fmt"
+	"go/types"
 	"reflect"
+	"slices"
 
 	"go.uber.org/nilaway/config"
 	"go.uber.org/nilaway/util/analysishelper"
 	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/types/objectpath"
 )
 
 const _doc = "Compute the per-package struct-field boundary summary: the field paths each function " +
@@ -34,6 +39,7 @@ var Analyzer = &analysis.Analyzer{
 	Doc:        _doc,
 	Run:        analysishelper.WrapRun(run),
 	ResultType: reflect.TypeOf((*analysishelper.Result[*ParamFieldEffects])(nil)),
+	FactTypes:  []analysis.Fact{new(ParamFieldReadsPackageFact)},
 	Requires:   []*analysis.Analyzer{config.Analyzer},
 }
 
@@ -43,5 +49,74 @@ func run(p *analysis.Pass) (*ParamFieldEffects, error) {
 	if !conf.ExperimentalStructInitV2Enable || !conf.IsPkgInScope(pass.Pkg) {
 		return &ParamFieldEffects{}, nil
 	}
-	return ComputeParamFieldEffects(pass), nil
+
+	packageSummary, forwardingEdges, callees := computeParamFieldEffects(pass)
+	if err := importUsedParamReads(pass, packageSummary.ParamReads, callees); err != nil {
+		return nil, err
+	}
+
+	closeParamFieldSets(packageSummary.ParamReads, forwardingEdges)
+	fact := &ParamFieldReadsPackageFact{}
+	encoder := &objectpath.Encoder{}
+	for funcObj := range packageSummary.ParamReads {
+		if funcObj.Pkg() != pass.Pkg || !funcObj.Exported() {
+			continue
+		}
+		// Export only parameter reads. Return reads remain local because callers infer them
+		// from their dereferences of results, opposite the direction of fact propagation.
+		reads := packageSummary.ParamReads.sortedPaths(funcObj)
+		if len(reads) == 0 {
+			continue
+		}
+		path, err := encoder.For(funcObj)
+		if err != nil {
+			return nil, fmt.Errorf("create object path for exported function %s: %w", funcObj, err)
+		}
+		fact.Functions = append(fact.Functions, FunctionParamFieldReads{FunctionObjectPath: path, ParamReads: reads})
+	}
+	if len(fact.Functions) > 0 {
+		// Functions were collected by ranging a map. Sorting makes the serialized fact
+		// deterministic and maintains the ordering required by BinarySearchFunc on import.
+		slices.SortFunc(fact.Functions, func(left, right FunctionParamFieldReads) int {
+			return cmp.Compare(left.FunctionObjectPath, right.FunctionObjectPath)
+		})
+		pass.ExportPackageFact(fact)
+	}
+
+	return packageSummary, nil
+}
+
+func importUsedParamReads(pass *analysishelper.EnhancedPass, paramReads fieldEffects, callees map[*types.Func]bool) error {
+	calleesByPackage := make(map[*types.Package][]*types.Func)
+	for callee := range callees {
+		if callee.Pkg() != nil && callee.Pkg() != pass.Pkg {
+			calleesByPackage[callee.Pkg()] = append(calleesByPackage[callee.Pkg()], callee)
+		}
+	}
+
+	packages := make([]*types.Package, 0, len(calleesByPackage))
+	for pkg := range calleesByPackage {
+		packages = append(packages, pkg)
+	}
+
+	encoder := &objectpath.Encoder{}
+	for _, pkg := range packages {
+		var fact ParamFieldReadsPackageFact
+		if !pass.ImportPackageFact(pkg, &fact) {
+			continue
+		}
+		for _, callee := range calleesByPackage[pkg] {
+			path, err := encoder.For(callee)
+			if err != nil {
+				return fmt.Errorf("create object path for imported function %s: %w", callee, err)
+			}
+			index, found := slices.BinarySearchFunc(fact.Functions, path, func(entry FunctionParamFieldReads, path objectpath.Path) int {
+				return cmp.Compare(entry.FunctionObjectPath, path)
+			})
+			if found {
+				seedImportedParamReads(paramReads, callee, fact.Functions[index].ParamReads)
+			}
+		}
+	}
+	return nil
 }
--- a/assertion/function/structfieldeffects/effects.go
+++ b/assertion/function/structfieldeffects/effects.go
@@ -27,8 +27,8 @@ import (
 	"golang.org/x/tools/go/types/typeutil"
 )
 
-// ParamFieldEffects is the package-level boundary summary computed once per package by
-// ComputeParamFieldEffects. Every effect set is keyed by *types.Func, then by indexedFieldPath.
+// ParamFieldEffects is the package-level boundary summary. Every effect set is keyed by
+// *types.Func, then by IndexedFieldPath.
 // The read sets bound the field binding at boundaries so it enumerates only the
 // field paths a boundary actually dereferences, never the full type graph.
 type ParamFieldEffects struct {
@@ -63,33 +63,35 @@ func readPaths(effects fieldEffects, funcObj *types.Func, idx int) []string {
 	reads := effects[funcObj]
 	paths := make([]string, 0, len(reads))
 	for key := range reads {
-		if key.idx == idx && key.path != "" {
-			paths = append(paths, key.path)
+		if key.Idx == idx && key.Path != "" {
+			paths = append(paths, key.Path)
 		}
 	}
+
+	// Sort paths so diagnostics at the same source location have a deterministic order.
 	sort.Strings(paths)
 	return paths
 }
 
-// ComputeParamFieldEffects walks every function and method in the package once and records the
-// boundary summary as ParamFieldEffects: the param fields it dereferences and the result fields
-// its callers dereference. It is a read-only, package-level pre-pass (pure
-// syntax/type inspection, no backpropagation).
+// computeParamFieldEffects walks every function and method in the package once and records the
+// unclosed boundary summary as ParamFieldEffects, plus forwarding edges for closure.
 //
 // Reads are gathered from selector bases — to evaluate `base.Sel`, base must be non-nil, so the
 // field path of base is a read of whatever boundary value it roots at (a parameter → ParamReads,
 // or a struct-returning-call result local → ReturnReads). ast.Inspect visits nested selectors, so
 // every prefix of a deep access is recorded. Every static call also records an arg→param forwarding
 // edge (which caller parameter, possibly at a nested field prefix, is passed as which callee
-// parameter). closeParamFieldSets then runs a fixpoint over those edges so forwarders inherit their
-// forwardees' param reads.
+// parameter). The analyzer later runs closeParamFieldSets over those edges so forwarders inherit
+// their forwardees' param reads. Static callees are returned so the analyzer imports only facts
+// needed by calls in this package.
 //
-// Cross-package and unresolvable (interface/func-value) callees contribute no edge and are treated
-// as mutating/dereferencing nothing (under-report only).
-func ComputeParamFieldEffects(pass *analysishelper.EnhancedPass) *ParamFieldEffects {
+// Unresolvable (interface/func-value) callees are treated as mutating/dereferencing nothing
+// (under-report only).
+func computeParamFieldEffects(pass *analysishelper.EnhancedPass) (*ParamFieldEffects, map[*types.Func][]paramFieldForwardEdge, map[*types.Func]bool) {
 	reads := make(fieldEffects)
 	returnReads := make(fieldEffects)
 	edges := make(map[*types.Func][]paramFieldForwardEdge)
+	callees := make(map[*types.Func]bool)
 	for _, file := range pass.Files {
 		for _, decl := range file.Decls {
 			fd, ok := decl.(*ast.FuncDecl)
@@ -117,7 +119,9 @@ func ComputeParamFieldEffects(pass *analysishelper.EnhancedPass) *ParamFieldEffe
 			ast.Inspect(fd.Body, func(n ast.Node) bool {
 				switch n := n.(type) {
 				case *ast.CallExpr:
-					collectParamForwardEdges(pass, n, paramIdx, funcObj, edges)
+					if callee := collectParamForwardEdges(pass, n, paramIdx, funcObj, edges); callee != nil {
+						callees[callee] = true
+					}
 				case *ast.SelectorExpr:
 					collectFieldReadDemand(pass, n, paramIdx, resultVars, funcObj, reads, returnReads)
 				}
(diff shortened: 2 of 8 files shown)


nilaway_test.go
```diff
@@ -116,6 +116,7 @@ func TestStructInitV2(t *testing.T) { //nolint:paralleltest
 		"go.uber.org/structinitv2/local",
 		"go.uber.org/structinitv2/defaultfield",
 		"go.uber.org/structinitv2/deep",
+		"go.uber.org/structinitv2/crosspkg/app",
```
yuxincs, line 119: Shouldn't we include the entire `crosspkg`?

assertion/function/structfieldeffects/fact_test.go
```diff
@@ -0,0 +1,222 @@
...
+	pass := results[0].Pass
+	exportedRead := pass.Pkg.Scope().Lookup("ExportedRead")
+	unusedRead := pass.Pkg.Scope().Lookup("UnusedRead")
+	require.NotNil(t, exportedRead)
+	require.NotNil(t, unusedRead)
+
+	fact := requirePackageFact(t, results[0].Facts[nil])
+	require.Len(t, fact.Functions, 2)
+	requireFunctionFact(t, fact, exportedRead.(*types.Func), expectedFactReadPaths())
+	requireFunctionFact(t, fact, unusedRead.(*types.Func), expectedFactReadPaths())
```
yuxincs: Question: can we not use golden tests for this (i.e., directly write `want` comments?)

assertion/function/structfieldeffects/fact.go
```diff
@@ -0,0 +1,31 @@
...
+type ParamFieldReadsPackageFact struct {
+	Functions []FunctionParamFieldReads
+}
+
+// AFact enables use of the facts passing mechanism in Go's analysis framework.
+func (*ParamFieldReadsPackageFact) AFact() {}
+
+// FunctionParamFieldReads is a function's parameter field-read summary within a package fact.
+type FunctionParamFieldReads struct {
+	ObjectPath objectpath.Path
```
yuxincs: It's not `ObjectPath`, it's specifically `Function`, right? it won't be any other object here.

assertion/function/structfieldeffects/effects.go
```diff
@@ -117,34 +117,63 @@ func ComputeParamFieldEffects(pass *analysishelper.EnhancedPass) *ParamFieldEffe
...
+			return paths[i].Idx < paths[j].Idx
+		}
+		return paths[i].Path < paths[j].Path
+	})
+	return paths
+}
+
+// IndexedFieldPath identifies a boundary value by parameter/result index and field path.
+// For example, in an access to `a.b.c` where `a` is the first parameter, {Idx: 0,
+// Path: "b"} represents the read demand on that parameter's `b` field.
```
yuxincs, line 167: random question: do we have to care about receivers?

```
func (a *Obj) Foo(b *Struct)
```

technically, `a` is the first index

assertion/function/structfieldeffects/fact_test.go
```diff
@@ -0,0 +1,222 @@
...
+	require.NotEmpty(t, buf.Bytes())
+	encoded := append([]byte(nil), buf.Bytes()...)
+
+	var decoded ParamFieldReadsPackageFact
+	require.NoError(t, gob.NewDecoder(&buf).Decode(&decoded))
+	require.Equal(t, fact, &decoded)
+
+	buf.Reset()
+	require.NoError(t, gob.NewEncoder(&buf).Encode(fact))
+	require.Equal(t, encoded, buf.Bytes(), "encoded fact must be deterministic")
```
yuxincs: Can we codec 10 times (check affilication analyzer or the inferred map tests) to ensure determinism?

assertion/function/structfieldeffects/fact_test.go
```diff
@@ -0,0 +1,222 @@
...
+			ObjectPath: "ExportedRead",
+			ParamReads: []IndexedFieldPath{
+				{Idx: -1, Path: "Mid"},
+				{Idx: 0, Path: "Mid.Child"},
+			},
+		},
+		{
+			ObjectPath: "Other",
+			ParamReads: []IndexedFieldPath{{Idx: 2, Path: "Other"}},
+		},
```
yuxincs: Can we generate a few more samples (in a loop or something) to better stress test our codec?

yuxincs: Can we also add a size check on the facts? this way if we updated logic in the future it broke our cap we would at least be aware (and make a determination to either increase the cap in the tests, or fix our logic)

assertion/function/structfieldeffects/fact_test.go
```diff
@@ -0,0 +1,222 @@
...
+func BenchmarkParamFieldReadsPackageFactCodec(b *testing.B) {
+	for _, functionCount := range []int{1, 10, 100} {
+		fact := newBenchmarkFact(functionCount)
+		b.Run(fmt.Sprintf("encode/%d-functions", functionCount), func(b *testing.B) {
+			b.ReportAllocs()
+			for b.Loop() {
+				var buf bytes.Buffer
+				if err := gob.NewEncoder(&buf).Encode(fact); err != nil {
+					b.Fatal(err)
+				}
```
yuxincs, line 93: Don't you need to create the encoder once and reuse that for benchmarking purposes?

assertion/function/structfieldeffects/fact_test.go
```diff
@@ -0,0 +1,222 @@
...
+	path, err := objectpath.For(funcObj)
+	require.NoError(t, err)
+	for _, function := range fact.Functions {
+		if function.ObjectPath == path {
+			require.Equal(t, want, function.ParamReads)
+			return
+		}
+	}
+	t.Fatalf("missing fact for %s", funcObj)
+}
```
yuxincs: These might not be necessary if we use golden tests, right?

assertion/function/structfieldeffects/analyzer.go
```diff
@@ -43,5 +48,72 @@ func run(p *analysis.Pass) (*ParamFieldEffects, error) {
...
+	closeParamFieldSets(packageSummary.ParamReads, forwardingEdges)
+	fact := &ParamFieldReadsPackageFact{}
+	encoder := &objectpath.Encoder{}
+	for funcObj := range packageSummary.ParamReads {
+		if funcObj.Pkg() != pass.Pkg || !funcObj.Exported() {
+			continue
+		}
+		// Export only parameter reads. Return reads remain local because callers infer them
+		// from their dereferences of results, opposite the direction of fact propagation.
+		reads := packageSummary.ParamReads.sortedPaths(funcObj)
```
yuxincs, line 67: IIUC, we would actually build the `ParamReads` by combining the local packages's fact and _all_ upstream packages' facts?

This would lead to fact explosions if the dependency chain is very long, right? should we just export the incremental part (i.e., facts that are only discovered for the local package)?

assertion/function/structfieldeffects/analyzer.go
```diff
@@ -43,5 +48,72 @@ func run(p *analysis.Pass) (*ParamFieldEffects, error) {
...
+	for _, pkg := range packages {
+		var fact ParamFieldReadsPackageFact
+		if !pass.ImportPackageFact(pkg, &fact) {
+			continue
+		}
+		for _, callee := range calleesByPackage[pkg] {
+			path, err := encoder.For(callee)
+			if err != nil {
+				continue
+			}
```
yuxincs, line 112: What kinds of errors can happen here? 

I'm not opposed to silently dropping the error, but we should know what kinds of errors we are dropping here.

assertion/function/structfieldeffects/fact.go
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package structfieldeffects
+
+import "golang.org/x/tools/go/types/objectpath"
+
+// ParamFieldReadsPackageFact is the exported parameter field-read summary for one package.
+type ParamFieldReadsPackageFact struct {
+	Functions []FunctionParamFieldReads
```
yuxincs, line 21: Question: if this is a slice, you actually need to do linear search if you'd like to access one function summary, right?

should we use a map instead? (we can always convert it to slice during encoding to preserve determinism)

assertion/function/structfieldeffects/effects.go
```diff
@@ -117,34 +117,63 @@ func ComputeParamFieldEffects(pass *analysishelper.EnhancedPass) *ParamFieldEffe
...
+	for _, read := range reads {
+		if read.Path == "" {
+			continue
+		}
+		paramReads.add(funcObj, read)
+	}
+}
+
+// sortedPaths returns funcObj's field paths in a deterministic order.
+func (e fieldEffects) sortedPaths(funcObj *types.Func) []IndexedFieldPath {
```
yuxincs, line 148: do we _have to_ sort it? 

I understand it's for having deterministic order, but within a single function, the order is already deterministic, right? (it would be the AST traversal order) it's just not sorted, would that suffice?

nilaway_test.go
karthikbhata97: Other packages under crosspkg are the "upstream packages" that is used by this package for the tests.

assertion/function/structfieldeffects/effects.go
karthikbhata97: Yes we do, we use the index -1 (ReceiverParamIndex const defined in annotation/key.go, this const predates struct-init-v2, so reused that).

assertion/function/structfieldeffects/analyzer.go
karthikbhata97: Yes, we are only exporting facts for the exported functions in the current package (`funcObj.Pkg() != pass.Pkg => continue`)

assertion/function/structfieldeffects/effects.go
karthikbhata97: The paths are stored as a set (i.e map[IndexedFieldPath]bool), so it does not have any order.

assertion/function/structfieldeffects/fact.go
karthikbhata97: The functions are kept in sorted order at the time of fact creation, so we perform binary search over it (slices.BinarySearchFuc) when looking up the function (analyzer.go:114)
We can add map, but i don't think it will make much of a difference

assertion/function/structfieldeffects/fact_test.go
karthikbhata97: I have been going back and forth with this one.
- fresh encoder / decoder: gives the actual overhead as it highlights the type registration cost (actual per package behavior)
- reusing encoder / decoder: only measures the actual fact gob encoding without type overhead 

So i think creating fresh encoder is more realistic benchmark. What do you think?

assertion/function/structfieldeffects/analyzer.go
karthikbhata97: Added error handling and return. 
The only error path as the object path encoder returning an error at `For`.

assertion/function/structfieldeffects/fact.go
karthikbhata97: Renamed to `FunctionObjectPath` to better reflect the intent

assertion/function/structfieldeffects/fact_test.go
karthikbhata97: Yes, that is better. Refactored.

assertion/function/structfieldeffects/fact_test.go
karthikbhata97: Updated

assertion/function/structfieldeffects/fact_test.go
karthikbhata97: Updated (1, 10 and 100 functions). Also added 10KB limit for a 100 function fact (4 param reads per function).

assertion/function/structfieldeffects/fact.go
yuxincs: OK, let's keep this as-is and optimize if this really brings performance hits

assertion/function/structfieldeffects/fact_test.go
yuxincs: Hmmmm, yes I agree that fresh encoder actually better captures the entire encoding :)

assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/upstream/upstream.go
```diff
@@ -0,0 +1,29 @@
...
+	Child *Leaf
+}
+
+type Outer struct {
+	Mid *Node
+}
+
+func ExportedRead(o *Outer) { // expect_reads: param_reads:0:Mid param_reads:0:Mid.Child
+	_ = o.Mid.Child.Ptr
+}
```
yuxincs, line 19: Could you clarify the meaning here?

we expect the param reads to be 

* 0-th parameter read on field `Mid`
* 0-th parameter read on field `Mid.Child`

and you don't need `.Ptr` since it's no longer a struct, right?

assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/upstream/upstream.go
```diff
@@ -0,0 +1,29 @@
...
+
+func ExportedNoRead(o *Outer) {
+	_ = o
+}
+
+func unexportedRead(o *Outer) { // expect_reads: param_reads:0:Mid param_reads:0:Mid.Child
+	_ = o.Mid.Child.Ptr
+}
```
yuxincs, line 33: Can we also have a case where we just read `o.Mid`?

assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/downstream/downstream.go
```diff
@@ -0,0 +1,15 @@
+package downstream // want package:".*"
+
+import "go.uber.org/paramfieldeffects/factexport/upstream"
+
+func Forward(o *upstream.Outer) { // expect_reads: param_reads:0:Mid param_reads:0:Mid.Child
```
yuxincs, line 5: This is because of forwarding, right?

assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/upstream/upstream.go
```diff
@@ -0,0 +1,29 @@
...
+
+type Outer struct {
+	Mid *Node
+}
+
+func ExportedRead(o *Outer) { // expect_reads: param_reads:0:Mid param_reads:0:Mid.Child
+	_ = o.Mid.Child.Ptr
+}
+
+func UnusedRead(o *Outer) { // expect_reads: param_reads:0:Mid param_reads:0:Mid.Child
```
yuxincs: why is this called `UnusedRead`? it is actually used, no?

assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/downstream/downstream.go
```diff
@@ -0,0 +1,15 @@
...
+	upstream.ExportedRead(o)
+}
+
+func ForwardUnused(o *upstream.Outer) { // expect_reads: param_reads:0:Mid param_reads:0:Mid.Child
+	upstream.UnusedRead(o)
```
yuxincs: Hmmm, I feel that you meant to test a case where it's forwarding `o` to an upstream function where it is not using `o` at all?

assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/upstream/upstream.go
karthikbhata97: .Ptr is not a demand because for a o.Mid.Child.Ptr, it only requires o, o.Mid, o.Mid.Child so field effects does not include demand for it.

assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/downstream/downstream.go
karthikbhata97: Yes

assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/downstream/downstream.go
karthikbhata97: Yes, fixed it. That was a mistake

assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/upstream/upstream.go
karthikbhata97: Also added a test for reading just the mid.

assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/upstream/upstream.go
karthikbhata97: It was a mistake, fixed it.

assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/upstream/upstream.go
karthikbhata97: Added `ReadOneLevelDeep`

yuxincs (approved):

PR #440, opened 2026-07-16
feat(struct-init-v2): propagate parameter field side effects

karthikbhata97:
Track field writes through function parameters and receivers so callers observe the field’s post-call nilability.

Changes
- Add parameter-output field contexts for direct and transitive writes.
- Reproduce written fields at call sites while preserving unaffected fields.
- Track forwarded writes through nested calls, receivers, and field paths.
- Enable parameter side-effect tests and document unsupported call-result composition.

### DIFF
--- a/assertion/function/assertiontree/root_assertion_node.go
+++ b/assertion/function/assertiontree/root_assertion_node.go
@@ -751,7 +751,13 @@ func (r *RootAssertionNode) AddComputation(expr ast.Expr) {
 			}
 
 			if r.functionContext.functionConfig.EnableStructInitV2 {
-				r.bindArgAndReceiverFieldsToContext(expr, fun)
+				funcObj := r.ObjectOf(fun).(*types.Func)
+				// First, let post-call field reads use the callee's param-out summary.
+				r.addCallParamOutFieldProducers(expr, funcObj)
+				// Then, if this function forwards its own parameter, make that summary its own output.
+				r.bindForwardedParamOut(expr, funcObj)
+				// Finally, bind the arguments' incoming fields to the callee's param-in summary.
+				r.bindArgAndReceiverFieldsToContext(expr, funcObj)
 			}
 		} else {
 			// here we have found either a builtin function like make or new,
--- a/testdata/src/go.uber.org/structinitv2/limitations/limitations.go
+++ b/testdata/src/go.uber.org/structinitv2/limitations/limitations.go
@@ -172,16 +172,6 @@ func tForwardViaFuncValue() *int {
 	return b.child.ptr
 }
 
-// False negative: parameter field writes are not supported, so a nil field written by a callee is
-// not re-produced at the call site.
-func clobberParamField(c *Node) { c.child = nil }
-
-func tParamFieldWriteNotMerged() *int {
-	b := &Node{child: &Leaf{}}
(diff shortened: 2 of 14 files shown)


testdata/src/go.uber.org/structinitv2/paramfield/paramfield.go
```diff
@@ -90,19 +90,6 @@ func f14(c *A) {
...
-}
-
-func callF15() {
-	f15(giveA15())
-}
-
-func f15(c *A) {
-	print(c.aptr.ptr) //want "field `aptr` of param 0 of `f15`"
-}
-
```
yuxincs, line 105: This and the following are added to limitations?

testdata/src/go.uber.org/structinitv2/limitations/limitations.go
```diff
@@ -172,16 +172,6 @@ func tForwardViaFuncValue() *int {
...
 
-// False negative: parameter field writes are not supported, so a nil field written by a callee is
-// not re-produced at the call site.
-func clobberParamField(c *Node) { c.child = nil }
-
-func tParamFieldWriteNotMerged() *int {
-	b := &Node{child: &Leaf{}}
-	clobberParamField(b)
-	return b.child.ptr
-}
```
yuxincs, line 183: should we add this back to the test suite?

assertion/function/assertiontree/root_assertion_node.go
```diff
@@ -751,7 +751,14 @@ func (r *RootAssertionNode) AddComputation(expr ast.Expr) {
...
 			if r.functionContext.functionConfig.EnableStructInitV2 {
-				r.bindArgAndReceiverFieldsToContext(expr, fun)
+				funcObj := r.ObjectOf(fun).(*types.Func)
+				sig := funcObj.Signature()
+				// First, let post-call field reads use the callee's param-out summary.
+				r.addCallParamOutFieldProducers(expr, funcObj, sig)
+				// Then, if this function forwards its own parameter, make that summary its own output.
+				r.bindForwardedParamOut(expr, funcObj, sig)
+				// Finally, bind the arguments' incoming fields to the callee's param-in summary.
+				r.bindArgAndReceiverFieldsToContext(expr, funcObj, sig)
```
yuxincs: `sig` can be simply drived from `obj.Signature` so we can probably just pass `funcObj`?

yuxincs (approved):

testdata/src/go.uber.org/structinitv2/paramfield/paramfield.go
karthikbhata97: Yes, these need the `return effect` changes as `f15` uses the return of `giveA15`. I have added them to limitations for now. Once the return changes are in I will bring them back

testdata/src/go.uber.org/structinitv2/limitations/limitations.go
karthikbhata97: Yes, this is added back (defaultfield.go)

assertion/function/assertiontree/root_assertion_node.go
karthikbhata97: Updated with this change

PR #442, opened 2026-07-17
feat(struct-init-v2): propagate parameter field writes across packages

karthikbhata97:
Propagate parameter-field write effects across package boundaries. This allows callers to observe direct, deep, and transitively forwarded mutations made by external callees.

Changes
- Extend the typed field-effects fact with parameter writes.
- Import read and write effects before transitive closure.
- Preserve deterministic fact serialization.
- Add cross-package write and forwarding coverage.
- Enable crosspkgside/app and paramfield tests.

### DIFF
--- a/assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/downstream/downstream.go
+++ b/assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/downstream/downstream.go
@@ -13,3 +13,19 @@ func ForwardUnused(o *upstream.Outer) {
 func DirectCall() {
 	upstream.ExportedRead(&upstream.Outer{})
 }
+
+func ForwardWrite(o *upstream.Outer) { // expect_effects: param_writes:0:Mid
+	upstream.ExportedWrite(o)
+}
+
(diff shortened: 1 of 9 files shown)


assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/downstream/downstream.go
```diff
@@ -13,3 +13,19 @@ func ForwardUnused(o *upstream.Outer) {
 func DirectCall() {
 	upstream.ExportedRead(&upstream.Outer{})
 }
+
+func ForwardWrite(o *upstream.Outer) { // expect_effects: writes:0:Mid
+	upstream.ExportedWrite(o)
+}
+
+func ForwardDeepWrite(o *upstream.Outer) { // expect_effects: param_reads:0:Mid writes:0:Mid.Child
```
yuxincs: `param_reads` and `writes`? perhaps we can keep the name consistent?

yuxincs (approved):

assertion/function/structfieldeffects/testdata/src/go.uber.org/paramfieldeffects/factexport/downstream/downstream.go
karthikbhata97: Updated to keep the consistency.

PR #446, opened 2026-07-21
feat(struct-init-v2): support generic static call field effects

karthikbhata97:
Enable field-effect bindings for statically resolved generic function and method calls.

- Add `typeshelper.StaticCallTarget`, separating a call's declared function origin from its instantiated call signature.
- Use the origin for field-effect/context-summary lookup and the instantiated signature for concrete argument, receiver, and result struct shapes.
- Handle direct generic calls such as `f[int]()` and instantiated generic methods.
- Add coverage for supported generic function and method returns.
- Document false-negative cases where a return shape depends on an unconstrained type parameter.

### DIFF
--- a/util/typeshelper/typeshelper.go
+++ b/util/typeshelper/typeshelper.go
@@ -17,10 +17,12 @@ package typeshelper
 
 import (
 	"fmt"
+	"go/ast"
 	"go/types"
 
 	"go.uber.org/nilaway/util/tokenhelper"
 	"golang.org/x/exp/typeparams"
+	"golang.org/x/tools/go/types/typeutil"
 )
 
 // ErrorType is the type of the builtin "error" interface.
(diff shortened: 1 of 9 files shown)


util/typeshelper/typeshelper.go
```diff
@@ -47,6 +49,36 @@ var BuiltinAppend = types.Universe.Lookup("append")
...
+type StaticCallTarget struct {
+	Origin    *types.Func
+	Signature *types.Signature
+}
+
+// GetStaticCallTarget returns the target of call when it has a statically resolved callee. For a
+// non-generic call, Origin identifies the declaration and Signature is its call signature. For a
+// generic instantiation, Origin still identifies the generic declaration while Signature includes
+// the type-argument substitutions.
+func GetStaticCallTarget(info *types.Info, call *ast.CallExpr) (StaticCallTarget, bool) {
```
yuxincs: We usually try to avoid `Get` in the name, so perhaps just name it `StaticCallTarget`

Also, as we discussed, let's try to upstream a fix such that we do not have to have a local workaround

yuxincs (approved):

util/typeshelper/typeshelper.go
karthikbhata97: `StaticCallTarget` is already the result struct type. So i renamed it to `ResolveStaticCallTarget`.

I have created an upsteram issue and PR (https://github.com/golang/go/issues/80533)

PR #447, opened 2026-07-21
feat(struct-init-v2): Support local variable return effects

karthikbhata97:
Support struct field-effect collection when a function returns a local variable.

Changes
- Track locals initialized from concrete struct allocations or struct-returning calls.
- Emit concrete return effects or return-forwarding edges when those locals are returned.
- Conservatively discard tracked locals after mutation, reassignment, aliasing, escape, address-taking, or closure capture.
- Add test coverage for allocation locals, forwarded results, multi-result calls, and invalidating operations.

### DIFF
--- a/assertion/function/structfieldeffects/effects.go
+++ b/assertion/function/structfieldeffects/effects.go
@@ -161,7 +161,7 @@ func (c *collectedFieldEffects) collectFunction(pass *analysishelper.EnhancedPas
 	// Locals bound directly to a struct-returning call, so a later dereference of the local's fields
 	// can be attributed to that callee's result (the return-read demand).
 	resultVars := collectStructResultVars(pass, fd.Body)
-	c.collectReturnEffects(pass, fd, funcObj)
+	c.collectReturnEffects(pass, fd, funcObj, resultVars)
 
 	ast.Inspect(fd.Body, func(n ast.Node) bool {
 		switch n := n.(type) {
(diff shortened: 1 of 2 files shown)


assertion/function/structfieldeffects/effects.go
```diff
@@ -356,11 +358,178 @@ func staticStructAllocation(pass *analysishelper.EnhancedPass, expr ast.Expr) (s
...
+			}
+		case *ast.ValueSpec:
+			if len(n.Names) != len(n.Values) {
+				return true
+			}
+			for i, ident := range n.Names {
+				record(ident, n.Values[i])
+			}
+		case *ast.FuncLit:
+			return false
```
yuxincs, line 396: anonymous functions could still allocate struct variable, right? this would be handled in the future?

assertion/function/structfieldeffects/effects.go
```diff
@@ -356,11 +358,178 @@ func staticStructAllocation(pass *analysishelper.EnhancedPass, expr ast.Expr) (s
...
+// precision and under-report instead of emitting an unsound effect.
+//
+// allocationVars and resultVars are the candidate sources: a var is unstable if it is a member of
+// either map and appears in a context that may reassign, mutate, or expose it.
+//
+// markNode looks for *uses* (TypesInfo.Uses) of candidate vars inside a subexpression. Uses — not
+// defs — are what matters here: a def simply introduces a new binding (e.g. `var s = ...` defining
+// `s`), whereas a use reads an existing candidate and is the signal that the candidate has been
+// aliased, observed, or flowed elsewhere.
+func collectUnstableStructVars(pass *analysishelper.EnhancedPass, body *ast.BlockStmt, allocationVars map[*types.Var]structAllocationSource, resultVars map[*types.Var]structResultSource) map[*types.Var]bool {
```
yuxincs: I worry that we're missing patterns here, instead of finding "unstable" patterns, should we instead conservatively find "stable" patterns?

yuxincs (approved):

PR #456, opened 2026-07-28
fix(struct-init-v2): bind param-out writes from any local-rooted value

karthikbhata97:
Bind param-out writes whose RHS is any local-rooted value — a local variable, a field chain rooted at one (`x.aptr = local.inner`), or the address of either (`x.aptr = &value`) — through the flow-sensitive consumption path. Previously only a bare local identifier qualified; the other forms fell back to a static snapshot that has no boundary annotation for a local, missing provably nil writes at post-call dereferences (false negative) and, for &value, losing the pointee's field shape (false positive).

Changes
- Extract the RHS classification into `isLocalRootedValue`, covering field chains and address-of; unwrap & since `ParseExprAsProducer` collapses &structValue to the pointee's trackable path.
- Add positive (nil field chain) and negative (address of local with initialized field) tests in structinitv2/paramsideeffect.

### DIFF
--- a/assertion/function/assertiontree/structinitv2.go
+++ b/assertion/function/assertiontree/structinitv2.go
@@ -147,6 +147,26 @@ func (r *RootAssertionNode) addFieldProducers(structType *types.Struct, fieldIni
 	}
 }
 
+// isLocalRootedValue reports whether expr is a function-local variable, a field chain rooted
+// at one, or the address of either. Locals have no boundary annotation to snapshot statically, so
+// callers must resolve them through their flow-sensitive producers instead. Address-of is
+// unwrapped because even a proven-non-nil pointer snapshot carries no pointee field shape.
+func (r *RootAssertionNode) isLocalRootedValue(expr ast.Expr) bool {
+	if u, ok := ast.Unparen(expr).(*ast.UnaryExpr); ok && u.Op == token.AND {
+		expr = u.X
+	}
+	base, _ := asthelper.SplitFieldChain(expr)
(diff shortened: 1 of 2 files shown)


assertion/function/assertiontree/structinitv2.go
```diff
@@ -147,6 +147,26 @@ func (r *RootAssertionNode) addFieldProducers(structType *types.Struct, fieldIni
...
 
+// isLocalRootedValue reports whether expr is a function-local variable, a field chain rooted
+// at one, or the address of either. Locals have no boundary annotation to snapshot statically, so
+// callers must resolve them through their flow-sensitive producers instead. Address-of is
+// unwrapped because even a proven-non-nil pointer snapshot carries no pointee field shape.
+func (r *RootAssertionNode) isLocalRootedValue(expr ast.Expr) bool {
+	if u, ok := ast.Unparen(expr).(*ast.UnaryExpr); ok && u.Op == token.AND {
+		expr = u.X
+	}
+	base, _ := asthelper.SplitFieldChain(expr)
```
yuxincs, line 158: I wonder if we should have another function that simply reports the base. We can then save some (frequent?) allocations.

yuxincs (approved):

assertion/function/assertiontree/structinitv2.go
karthikbhata97: Hmm, I'll take note and measure if there is significant value in having just this functionality. If so, i can create that refactor.

PR #459, opened 2026-07-29
fix(inference): register missing gob types encoded in exported facts

karthikbhata97:
Register the concrete `Repr` types that can be encoded in exported `InferredMap` facts, preventing modular drivers from failing at fact-export time with `gob: type not registered for interface`.

Changes
- Register `BlankVarReturnRepr`, `ConsumeTriggerTautologyRepr`, `TriggerIfNonNilRepr`, and `TriggerIfDeepNonNilRepr`.
- Fix `ArgPassDeep.Repr()` to return its intended `ArgPassDeepRepr`, preserving the distinct deep-flow diagnostic and its gob registration.
- Add `TestGobRegistrationCompleteness`, which discovers fact interface slots and SSA conversions and diffs their concrete types against `gobRegisteredTypes` in both directions.

This PR builds on #458, which makes the registration list inspectable.

### DIFF
--- a/inference/gob_register_test.go
+++ b/inference/gob_register_test.go
@@ -1,0 +1,389 @@
+//  Copyright (c) 2026 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package inference
+
+import (
+	"go/types"
+	"reflect"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/tools/go/packages"
+	"golang.org/x/tools/go/ssa"
+	"golang.org/x/tools/go/ssa/ssautil"
+)
+
+const (
+	// _modulePattern loads every package in the NilAway module: facts and the types stored inside
+	// them may be defined anywhere in the module.
+	_modulePattern = "go.uber.org/nilaway/..."
+)
+
+// gobCandidate describes a concrete type that may be stored in an interface-typed slot of a
+// gob-encoded fact, and hence must be registered with gob.
+type gobCandidate struct {
+	obj *types.TypeName
+	// ptr indicates that the concrete value converted to the interface is a pointer.
+	ptr bool
+}
+
+// TestGobRegistrationCompleteness statically verifies that every concrete type that may be gob
+// encoded as part of an exported fact is registered with gob. Forgetting to register a type does
+// _not_ fail in-process drivers (e.g., our own analysistest-based tests or the standalone
+// checker, which pass facts in memory), but crashes modular drivers that serialize facts via gob
+// (go vet -vettool, bazel/nogo, golangci-lint) at fact-export time with
+// "gob: type not registered for interface: ...". This test lets CI catch such gaps early.
+//
+// Note that the fact types themselves (e.g., *InferredMap) never require manual registration:
+// analysis drivers register everything declared in Analyzer.FactTypes automatically. Manual
+// registration is only required for the dynamic types stored in interface-typed slots _inside_ a
+// fact, which is precisely what this test verifies.
+//
+// The test works as follows:
+//  1. It discovers every fact type in the module (i.e., types with an AFact method) and walks the
+//     type graph that gob would encode for each of them (exported fields only; for fact types
+//     with a custom GobEncode method, the types actually passed to (*gob.Encoder).Encode inside
+//     that method), and collects every interface-typed slot it encounters.
+//  2. It builds SSA for the module and, for each such interface, collects the concrete type from
+//     every conversion to that interface. These conversions are the complete set of ways concrete
+//     values enter the interface; this avoids guessing from type names or including types that
+//     implement an interface but are never stored in it.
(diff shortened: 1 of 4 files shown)


inference/gob_register_test.go
```diff
@@ -0,0 +1,474 @@
...
+	// and must be updated, since we can no longer trust the completeness of the candidate set.
+	for _, known := range []string{
+		_inferencePkgPath + ".InferredVal",
+		_inferencePkgPath + ".ExplainedBool",
+		"fmt.Stringer",
+	} {
+		require.Contains(t, ifaceNames, known,
+			"traversal of the facts did not discover interface %s; the traversal logic in "+
+				"this test likely needs updating", known)
+	}
```
yuxincs: This can probably be removed since it's checking a bit too rigorously?

inference/gob_register_test.go
```diff
@@ -0,0 +1,474 @@
...
+			stale = append(stale, "  - "+key)
+		}
+	}
+	sort.Strings(stale)
+	if len(stale) > 0 {
+		t.Errorf("%d type(s) in gobRegisteredTypes were not discovered as candidates:\n%s\n"+
+			"Either the registrations are dead and can be removed from gobRegisteredTypes "+
+			"(inference/engine.go), or the discovery logic in this test needs updating.",
+			len(stale), strings.Join(stale, "\n"))
+	}
```
yuxincs: You can use [cmp.Diff](https://github.com/google/go-cmp) such that you don't have to reinvent it :)

inference/gob_register_test.go
```diff
@@ -0,0 +1,474 @@
...
+func (c gobCandidate) registrationExpr() string {
+	name := c.obj.Name()
+	if c.obj.Pkg().Path() != _inferencePkgPath {
+		name = c.obj.Pkg().Name() + "." + name
+	}
+	if c.ptr {
+		return "&" + name + "{}"
+	}
+	return name + "{}"
+}
```
yuxincs, line 43: Let's move the helper functions after the main `Test*` functions

inference/gob_register_test.go
```diff
@@ -0,0 +1,474 @@
...
+			if !methods["AFact"] {
+				continue
+			}
+			factName := pkg.PkgPath + "." + obj.Name()
+			factNames = append(factNames, factName)
+
+			if !methods["GobEncode"] {
+				// Standard gob encoding: traverse the fact type itself (the traversal follows
+				// exported fields only, matching gob semantics).
+				roots = append(roots, named)
```
yuxincs, line 160: I wonder why we can't simply delve deeper and collect all fact names here. It would be better encapsulation than letting upper-level handle it, right?

inference/gob_register_test.go
```diff
@@ -0,0 +1,474 @@
...
+func gobEncodeRoots(pkg *packages.Package, factType *types.Named) []types.Type {
+	var roots []types.Type
+	for _, file := range pkg.Syntax {
+		ast.Inspect(file, func(n ast.Node) bool {
+			fd, ok := n.(*ast.FuncDecl)
+			if !ok || fd.Name.Name != "GobEncode" || fd.Recv == nil || fd.Body == nil {
+				return true
+			}
+			// Ensure this GobEncode is the method of the fact type under consideration.
+			fn, ok := pkg.TypesInfo.Defs[fd.Name].(*types.Func)
```
yuxincs: Use `.ObjectOf()`?

inference/gob_register_test.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+// gobEncodeRoots discovers the type graph roots of a fact with a custom GobEncode method: the
+// static types of all values passed to (*gob.Encoder).Encode within that method's body. This is
+// exactly what gob sees at encoding time, without needing hardcoded knowledge of which fields the
+// encoder chooses to serialize.
+func gobEncodeRoots(pkg *packages.Package, factType *types.Named) []types.Type {
+	var roots []types.Type
+	for _, file := range pkg.Syntax {
+		ast.Inspect(file, func(n ast.Node) bool {
```
yuxincs: Why do we have to start from package level again? previously we were already looking at the `factType`, can we simply obtain the function and scan from there?

inference/gob_register_test.go
```diff
@@ -0,0 +1,474 @@
...
+	return roots
+}
+
+// collectGobCandidates walks the gob-encoded type graph starting from the given fact roots and
+// returns the concrete types that may be stored in interface-typed slots, along with the set of
+// discovered interface names (for sanity checking).
+func collectGobCandidates(
+	t *testing.T, conversions []interfaceConversion, roots []types.Type,
+) ([]gobCandidate, map[string]bool) {
+	var (
```
yuxincs, line 220: call `t.Helper()` at the beginning of the helper such that the error messages are properly attributed to the upper test that calls this

inference/gob_register_test.go
```diff
@@ -0,0 +1,474 @@
...
+		case *types.Interface:
+			// Unnamed interfaces (e.g., `any`) cannot be handled by this test.
+			require.Fail(t, "unnamed interface type discovered in the gob-encoded type graph",
+				"type %s cannot be enumerated; please restructure the fact types or extend this test", typ)
+		default:
+			require.Fail(t, "unsupported type discovered in the gob-encoded type graph",
+				"type %s (%T) is not supported by this test; please extend the traversal logic", typ, typ)
+		}
+	}
+	return candidates, ifaceNames
```
yuxincs: it doesn't seem that we are using `ifaceNames` in meaningful ways

nilaway_test.go
```diff
@@ -76,6 +76,7 @@ func TestNilAway(t *testing.T) {
 		{name: "NoLint", patterns: []string{"go.uber.org/nolint"}},
 		{name: "Templ", patterns: []string{"go.uber.org/templ"}},
 		{name: "CtrlflowIntrinsic", patterns: []string{"go.uber.org/zap"}},
+		{name: "BlankVarReturnFactEncoding", patterns: []string{"go.uber.org/blankvarreturn"}},
```
yuxincs: Is it possible to blend this in one of the tests we already have? creating a different test suite just for encoding seems a bit excessive

yuxincs (approved):

PR #467, opened 2026-07-30
Remove more no inference markers

yuxincs:
Continuing on our previous endeaver, we remove more no inference markers in test packages in this PR


testdata/src/go.uber.org/arrays/arrays.go
```diff
@@ -30,42 +28,60 @@ func testArrayRet() [2]*int {
 	return globalArr
 }
 
-// nilable(v)
-// nonnil(a[])
 func testParamArrayWrite(a [4]*int, v *int, b bool) (*int, *int) {
```
karthikbhata97: Do we need these args if we are not going to use them?

karthikbhata97: Similarly in other test cases as well.

yuxincs: Nice catch, lemme fix!

karthikbhata97 (approved):

PR #469, opened 2026-07-30
Remove all remaining no inference markers in test packages

yuxincs:
This PR removes all remaining `<nilaway no inference>` markers in test code, after which we can finally deprecate NoInfer mode.

All annotation-mode test packages (`channels`, `contracts`, `deepnil`, `errorreturn`, `maps`, `methodimplementation`, `multipleassignment`, `namedreturn`, `receivers`) are rewritten so that nilability is established purely by code and inferred by NilAway — no `nilable(...)` / `nonnil(...)` annotations remain outside `annotationparse`, which keeps them intentionally to test annotation parsing itself under inference. The former `inference/` subpackages are promoted into the main packages (their coverage is now the default), the `-with-inference` file suffixes are dropped since everything is tested with inference now (`alwayssafe.go`, `deepnil2.go`, `errorreturn2.go`, `receivers2.go`, `downstream.go`), and cross-package coverage (e.g. the `otherPkg.Wrapf` error-wrapper pair) is preserved.

One gap even closed along the way: ranging over a channel *parameter* does produce a deep nil flow when the value is dereferenced in place (the previously observed silence was specific to helper-argpass consumption), so that case is now positively covered.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


testdata/src/go.uber.org/annotationparse/annotationparse.go
```diff
@@ -135,62 +154,76 @@ type (
...
-		return b2 //want "returned"
+		return b2
 	default:
 		m1.a = nil
-		m1.b = nil //want "assigned into field"
-		m2.a = nil //want "assigned into field"
+		m1.b = nil
+		print(*m1.b) //want "dereferenced"
+		m2.a = nil
+		print(*m2.a) //want "dereferenced"
```
karthikbhata97: This is not necessarily annotation parse behavior since it looks more of an inference test.
Should these tests be narrowed down?

testdata/src/go.uber.org/deepnil/deepnil.go
```diff
@@ -13,336 +13,419 @@
...
+	_ = *i //want "passed"
+}
 
 var i = 0
 
+// testsArrRets checks reads (in various index / local variable / range forms) of the deeply
+// nilable and deeply nonnil results of the two functions above. Since a single error is reported
+// per inference site, only one read of `retsNilableArr`'s result is dereferenced (case 1); the
+// other nilable reads are returned without dereference and are caught (as a single grouped error)
+// at the dereference of `testsArrRets`'s result in callsTestsArrRets below. Reads of
```
karthikbhata97: Since different cases also call functions other than `retsNilableArr`, wouldn't this result in failing to catch regressions, especially when one of the behaviors change the from returning nilable to nonnil but since we are only checking the shared return site at the caller, because of other arms providing nilable, it still passes.
Or if the other cases are already tested separately, maybe they are redundant here?

testdata/src/go.uber.org/annotationparse/annotationparse.go
yuxincs: Yeah I agree, but it's not possible to test this behavior right now. Because we are suppressing the field assignment triggers in full inference mode. This serves as a workaround to at least test some inference. I have added a TODO comment here.

testdata/src/go.uber.org/deepnil/deepnil.go
yuxincs: Yes! good catch! 

I have broken this into independent cases and removed some redundant cases that are already tested elsewhere.

karthikbhata97 (approved):

PR #472, opened 2026-07-30
feat(struct-init-v2): collect return param sources

karthikbhata97:
Collect, per function, the relations `result (field) <- parameter (field)` that make a returned value depend on the caller's own argument — the marks later revisions use to sever caller-dependent results from the shared return summary and resolve them per call site. Collection is syntactic and refuses anything conditional or ambiguous; disagreeing sources for one result are all kept so consumers can detect the disagreement and refuse to answer. Collector only: sources stay package-internal and nothing consumes them yet, so reported errors are unchanged.

Changes
- Record whole-result sources at return sites (`return p`, `return p.x`, receiver variants) in `collectWholeResultParamSource`; spread returns record nothing.
- Record field-level sources for construction fields initialized from a parameter or parameter field chain (`return &T{f: p.x}`) in `enumerateConcreteReturnEffects`, for any field kind: a source on a value-struct field re-roots deeper accesses of the result under the parameter.
- Never source unstable parameters: reassigned, address-taken, or closure-reassigned parameters may not hold the caller's argument at the return (`collectUnstableParamVars`).
- Drop the whole-result sources of mixed construct/forward results in `close()`: a result with both an in-package construction return site (`resultsWithConstructSite`) and a whole-result source has no context-insensitive owner (`dropMixedResultParamSources`).
- Gate collected paths with the existing acyclicity checks on both the result and parameter sides.
- Rename the return walk to `collectReturnSites` / `collectReturnSiteEffect`: one loop owns the site guards so the param-source and effect collections see identical return sites — the mixed drop relies on that agreement.
- Cover with collector fixtures in `returneffects/paramdependent.go` (`return_param_source` expectations for constructor fields, deep chains, receiver projections, and every drop rule).

### DIFF
--- a/assertion/function/structfieldeffects/effects_test.go
+++ b/assertion/function/structfieldeffects/effects_test.go
@@ -30,7 +30,9 @@ import (
 
 // _expectEffectsPrefix precedes a fixture function's expected boundary effects. Each trailing token
 // is "<kind>:<idx>:<path>", where kind identifies parameter writes, parameter reads, return reads,
-// or concrete return effects. A function with no effects carries an empty comment.
+// or concrete return effects; return param sources use the five-part
+// "return_param_source:<resultIdx>:<resultPath>:<paramIdx>:<paramPath>". A function with no effects
+// carries an empty comment.
 const _expectEffectsPrefix = "expect_effects:"
 
 func TestComputeBoundaryFieldEffects(t *testing.T) {
@@ -59,12 +61,17 @@ func TestComputeBoundaryFieldEffects(t *testing.T) {
 		wantReturn := make(map[*types.Func][]IndexedFieldPath)
 		wantWrites := make(map[*types.Func][]IndexedFieldPath)
 		wantReturnEffects := make(map[*types.Func][]IndexedFieldPath)
+		wantSources := make(map[*types.Func][]ReturnParamSource)
 		for node, tokens := range nilawaytest.FindExpectedValues(pass, _expectEffectsPrefix) {
 			fd, ok := node.(*ast.FuncDecl)
 			require.True(t, ok)
 			funcObj, ok := pass.TypesInfo.ObjectOf(fd.Name).(*types.Func)
 			require.True(t, ok)
 			for _, token := range tokens {
+				if source, ok := parseExpectedParamSource(t, token); ok {
+					wantSources[funcObj] = append(wantSources[funcObj], source)
+					continue
+				}
 				kind, key := parseExpectedEffect(t, token)
 				switch kind {
 				case "param_writes":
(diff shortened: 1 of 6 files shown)


assertion/function/structfieldeffects/return_contracts.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+		if _, isParam := fc.paramIdx[v]; !isParam {
+			return
+		}
+		if out == nil {
+			out = make(map[*types.Var]bool)
+		}
+		out[v] = true
+	}
+	ast.Inspect(fc.fd.Body, func(n ast.Node) bool {
```
yuxincs, line 188: Oh, since we have so many `ast.Inspect` in NilAway now, let's see if we can replace all of them with https://pkg.go.dev/golang.org/x/tools/go/ast/inspector such that we can walk it once and then use the results.

I'm not sure if we can directly use https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/inspect such that we get the construction for free (i.e., since it would be initialized anyways for other analyzers), since during our preprocessing we sometimes inject fake AST nodes.

If we can't, I guess we can always have a dedicated NilAway analyzer that constructs this once (after the preprocessing), and then later we can all just use the results.

This may yield some performance gains, once we get to the optimization phase.

Of course, that should be separate PRs.

assertion/function/structfieldeffects/effects_test.go
```diff
@@ -85,7 +92,26 @@ func TestComputeBoundaryFieldEffects(t *testing.T) {
 		requireEffects(t, effects.ReturnReads, wantReturn)
 		requireEffects(t, effects.ParamWrites, wantWrites)
 		requireEffects(t, effects.ReturnEffects, wantReturnEffects)
+		requireParamSources(t, effects, wantSources)
+	}
+}
+
+// parseExpectedParamSource parses a "return_param_source:<resultIdx>:<resultPath>:<paramIdx>:<paramPath>"
+// token, reporting ok=false for tokens of any other kind.
+func parseExpectedParamSource(t *testing.T, token string) (ReturnParamSource, bool) {
```
yuxincs, line 101: call `t.Helper()` such that error would be properly aggregated under the caller

assertion/function/structfieldeffects/effects_test.go
```diff
@@ -115,3 +141,19 @@ func requireEffects(t *testing.T, got fieldEffects, want map[*types.Func][]Index
 		require.ElementsMatchf(t, want[funcObj], gotKeys, "effects mismatch for %s", funcObj.Name())
 	}
 }
+
+// requireParamSources asserts the collected return param sources match want for every function in either
+// set, so both missing and unexpected sources fail the test.
+func requireParamSources(t *testing.T, got *BoundaryFieldEffects, want map[*types.Func][]ReturnParamSource) {
```
yuxincs, line 148: call `t.Helper()`

yuxincs (approved):

assertion/function/structfieldeffects/effects_test.go
karthikbhata97: Fixed

assertion/function/structfieldeffects/effects_test.go
karthikbhata97: Fixed

PR #475, opened 2026-08-01
Validate codecs for all exported NilAway facts

yuxincs:
This change validates gob encoding for every fact produced by NilAway analyzers. It verifies deterministic encoding, successful decoding, byte-identical round trips, and aggregate fact-size limits. It also removes the test-only production check and redundant InferredMap codec tests.

This ensures much more thorough testing for NilAway exported facts rather than inferred map's synthetic codec tests, and will catch fact size problems in an easier way.


nilaway_test.go
```diff
@@ -124,6 +140,13 @@ func TestStructInitV2(t *testing.T) { //nolint:paralleltest
 		"go.uber.org/structinitv2/returnlocal",
 		"go.uber.org/structinitv2/returnzerovalue/app",
 	)
+	factStats := nilawaytest.RequireFactCodecs(t, results)
+	t.Logf("Total fact bytes: %d", factStats.TotalBytes)
+	require.Positive(t, factStats.TotalBytes, "expected struct initialization v2 tests to export some facts")
+	require.Less(t, factStats.TotalBytes, 25*1024,
+		"The gob encoded NilAway facts for all struct initialization v2 test code is too large (%.2f KB). "+
+			"We expect the total to be less than 25KB. This heavily affects the artifact sizes of the facts NilAway "+
+			"produces, so the cap should only be increased with justification and thorough testing.", float32(factStats.TotalBytes)/1024)
```
yuxincs: Adding a cap here for now, this would eventually be folded into the main TestNilAway.

sonalmahajan15 (approved):

PR #478, opened 2026-08-02
fix(struct-init-v2): keep address-of field writes non-nil at the pointer level

karthikbhata97:
Address-of assignments such as `d.resp = &mr` incorrectly re-keyed the destination pointer obligation onto the pointee producer in both intra-procedural backpropagation and parameter-output summaries. That produced a false positive when `mr` came from a guarded result even though `&mr` is always non-nil; preserving the address expression as a never-nil shallow producer removes the pointer-level error while retaining pointee field flows.

Changes
- Match direct LHS consumers with a never-nil producer before the case-C landing so child obligations continue through the pointee and still report nil fields.
- Exclude address-of expressions from `isLocalRootedValue` so `bindParamFieldWriteToContext` snapshots their always-non-nil shallow producer instead of associating the `PARAM_OUT` supply with the pointee.
- Add `paramsideeffect` coverage for store-before-check flows, caller reads, nil pointee fields, genuinely nil writes, local aliases, field chains, and nilable pointer results.

### DIFF
--- a/assertion/function/assertiontree/backprop.go
+++ b/assertion/function/assertiontree/backprop.go
@@ -714,6 +714,23 @@ buildShadowMask:
 							}
 						}
 
+						if rootNode.functionContext.functionConfig.EnableStructInitV2 {
+							if u, ok := ast.Unparen(rhsVal).(*ast.UnaryExpr); ok && u.Op == token.AND {
+								// The address is non-nil, while its fields still land on the pointee below.
+								producer := &annotation.ProduceTrigger{
+									Annotation: &annotation.ProduceTriggerNever{},
+									Expr:       rhsVal,
+								}
+								for _, consumer := range lhsNode.ConsumeTriggers() {
+									rootNode.AddNewTriggers(annotation.FullTrigger{
+										Producer: producer,
+										Consumer: consumer,
+									})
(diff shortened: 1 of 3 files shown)


assertion/function/assertiontree/backprop.go
```diff
@@ -714,6 +714,23 @@ buildShadowMask:
 							}
 						}
 
+						if rootNode.functionContext.functionConfig.EnableStructInitV2 {
```
yuxincs, line 717: Any reason this has to be guarded under the flag? Seems like a general fix? Or did I misunderstand the problem?

karthikbhata97: It is already done for the general case at https://github.com/uber-go/nilaway/blob/e5f776ef57e1d0e0d406e7fcbcee34b9a7c5bbe5/assertion/function/assertiontree/backprop.go#L327-L340

But for struct-init-v2, it is not done there since `AddProduction` there will detach the LHS subtree.

So it is delayed until later.

yuxincs: Ah that sounds good! Thx!

yuxincs (approved):

PR #482, opened 2026-08-03
Fail golden tests on internal panics

yuxincs:
- Fail golden tests when either branch reports an `INTERNAL PANIC` diagnostic.
- Print internal panic diagnostics first so they remain visible in truncated output.

### DIFF
--- a/accumulation/analyzer.go
+++ b/accumulation/analyzer.go
@@ -83,7 +83,10 @@ func run(p *analysis.Pass) (result interface{}, _ error) {
 			// Deferred functions are executed after a result is generated, so here we modify the
 			// return value `result` in-place.
 			// Diagnostics with invalid positions (<= 0) will be silently suppressed, so here we use 1.
-			d := analysis.Diagnostic{Pos: 1, Message: fmt.Sprintf("INTERNAL PANIC: %s\n%s", r, string(debug.Stack()))}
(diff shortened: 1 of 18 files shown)


accumulation/analyzer.go
```diff
@@ -83,7 +83,10 @@ func run(p *analysis.Pass) (result interface{}, _ error) {
 			// Deferred functions are executed after a result is generated, so here we modify the
 			// return value `result` in-place.
 			// Diagnostics with invalid positions (<= 0) will be silently suppressed, so here we use 1.
-			d := analysis.Diagnostic{Pos: 1, Message: fmt.Sprintf("INTERNAL PANIC: %s\n%s", r, string(debug.Stack()))}
+			d := analysis.Diagnostic{
+				Pos:     1,
+				Message: fmt.Sprintf("%s: %s\n%s", config.InternalPanicPrefix, r, string(debug.Stack())),
```
karthikbhata97, line 88: nit: Curious if this can be encoded as a member `Category` so we can skip the substring match.

karthikbhata97 (approved):

accumulation/analyzer.go
yuxincs: yeah good call! I don't actually recall why we didn't leverage the Category field, maybe it's time to revisit this

PR #490, opened 2026-08-04
feat(struct-init-v2): resolve parameter-dependent shallow results per call site

karthikbhata97:
Resolve that root demand through the same call-site source and site so shallow nilability remains caller-specific.

Changes
- Bind root result demands before param-out productions so shallow results observe post-call argument values.
- Read tracked and untracked call values from the located return site instead of the shared declaration return.
- Keep identical calls at different locations distinct and resolve projected fields of inline allocation arguments.
- Cover direct function, method, projection, and escaped-field shallow-result scenarios.

### DIFF
--- a/assertion/function/assertiontree/func_assertion_node.go
+++ b/assertion/function/assertiontree/func_assertion_node.go
@@ -29,6 +29,10 @@ type funcAssertionNode struct {
 	// declaring identifier for this function
 	decl *types.Func
 	args []ast.Expr
+	// call identifies the source call. With -experimental-struct-init-v2 enabled, shallowEqNodes
+	// uses its position to distinguish nodes only when the result has a unique whole-result
+	// parameter source.
+	call *ast.CallExpr
 }
 
 func (f *funcAssertionNode) MinimalString() string {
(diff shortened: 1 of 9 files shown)


assertion/function/assertiontree/func_assertion_node.go
```diff
@@ -29,6 +29,8 @@ type funcAssertionNode struct {
 	// declaring identifier for this function
 	decl *types.Func
 	args []ast.Expr
+	// call identifies the site of a parameter-sourced result.
+	call *ast.CallExpr
```
yuxincs: Would this actually incur a lot of performance penalty? because we are practically adding context sensitivity for all cases, right? meaning each call would result in a different func assertion node

karthikbhata97: https://github.com/uber-go/nilaway/blob/4abcedfcb195d38e4c870731e8a4b255fb1000e5/assertion/function/assertiontree/root_assertion_node.go#L1129-L1136

Here the `shallowEqNodes` consume `call` only on `struct init v2` and there is param dependent return. At other places, it's just an extra field.

yuxincs (approved):

PR #494, opened 2026-08-05
Remove no-inference mode

yuxincs:
This change removes NilAway’s legacy no-inference mode and makes inference the only analysis path. It deletes mode-selection logic, the package directive, obsolete annotation-checking APIs, and related dead code. It also migrates affected testdata and stubs to inference mode and updates stale comments.


inference/engine.go
```diff
@@ -128,22 +128,17 @@ func (e *Engine) ObserveUpstream() {
...
+func (e *Engine) ObserveAnnotations(pkgAnnotations *annotation.ObservedMap) {
 	pkgAnnotations.Range(func(key annotation.Key, isDeep bool, val bool) {
 		site := e.primitive.site(key, isDeep)
 		if val {
 			e.observeSiteExplanation(site, TrueBecauseAnnotation{AnnotationPos: site.Position})
 		} else {
 			e.observeSiteExplanation(site, FalseBecauseAnnotation{AnnotationPos: site.Position})
 		}
-	}, mode != NoInfer)
+	}, true)
```
karthikbhata97: Probably we can drop that second argument from the definition, if this is the only usecase.

yuxincs: Good catch!! removed the redundant arg.

karthikbhata97 (approved):

PR #496, opened 2026-08-05
Run integration tests against the full analysistest corpus

yuxincs:
This change redirects integration tests to testdata/src/go.uber.org, providing broader coverage across 1,031 expected diagnostics. It also removes the obsolete integration-test corpus and improves diagnostic matching.


.codecov.yaml
```diff
@@ -1,3 +1,6 @@
+ignore:
+  - "tools"
+
```
yuxincs, line 3: Bundling this change (we should've done this before) such that codecov won't prevent this PR from passing

tools/cmd/integration-test/main.go
```diff
@@ -41,75 +44,216 @@ type Position struct {
...
+func CollectGroundTruths(dir string) (map[Position][]*regexp.Regexp, error) {
+	truths := make(map[Position][]*regexp.Regexp)
+	fset := token.NewFileSet()
+	err := filepath.WalkDir(dir, func(path string, entry fs.DirEntry, walkErr error) error {
+		if walkErr != nil {
+			return walkErr
+		}
+		if entry.IsDir() || filepath.Ext(path) != ".go" {
+			return nil
+		}
```
yuxincs, line 60: Converting from "loading the package and then get the AST" to simply parse the files without loading the packages. It is faster, and it avoid package loading errors for unit testdata (since they don't really have go.mod files, which the package loader requires)

tools/cmd/integration-test/golangci_lint.go
```diff
@@ -81,7 +81,16 @@ func (d *GolangCILintDriver) Run(dir string) (diagnostics map[Position]string, e
...
 	diagnosticFile := filepath.Join(tempDir, "diagnostics.json")
-	cmd = exec.Command(filepath.Join(tempDir, "custom-gcl"), "run", "--output.json.path", diagnosticFile, "--path-mode", "abs", "./...")
+	cmd = exec.Command(
+		filepath.Join(tempDir, "custom-gcl"),
+		"run",
+		"--output.json.path", diagnosticFile,
+		"--path-mode", "abs",
+		"--max-issues-per-linter=0",
+		"--max-same-issues=0",
+		"--uniq-by-line=false",
```
yuxincs, line 91: Has to be relaxed in order to report all diagnostics

tools/cmd/integration-test/main.go
```diff
@@ -41,75 +44,175 @@ type Position struct {
 type Driver interface {
 	// Run runs NilAway on the test project specified by dir and returns the diagnostics reported
 	// by NilAway (in a map from Position to the diagnostic message).
-	Run(dir string) (map[Position]string, error)
+	Run(dir string) (map[Position][]string, error)
```
yuxincs, line 47: We have to make this return a list of diagnostics because in `testdata/go.uber.org` there might be multiple diagostics reported on a single location

testdata/integration/contracts/downstream/downstream.go
```diff
@@ -1,33 +0,0 @@
-package downstream
```
yuxincs, line 1: All adhoc integration test code is now removed (and some harvested to unit tests)

testdata/src/go.uber.org/annotationparse/annotationparse.go
```diff
@@ -48,7 +48,7 @@ func foo(a, b *bar, c *bar, d, e *bar) (f, g *bar, h *bar) { //want "returned fr
 	myBar.karp = a
 	myBar.lug = a
 	myBar.myr = a
-	print(*myBar.myr) //want "dereferenced"
+	_ = *myBar.myr //want "dereferenced"
```
yuxincs, line 51: The previous `print` actually caused compilation errors (somehow `analysistest` allowed it). When running against the integration tests (that use real drivers), it would fail.

So, here we convert them to simply dereference.s

karthikbhata97 (approved):

PR #499, opened 2026-08-05
fix(map-read): guard only maps with nilable values

karthikbhata97:
Avoid requiring missing-key guards for maps whose element type cannot be nil.

Changes
- Add generic `IsDeeplyTypeWithNilableElement[T]` for named, aliased, and constrained types.
- Use it at deep map producers so value-typed map reads do not emit false positives.
- Keep grouping coverage guard-sensitive and cover map and slice instantiations.

### DIFF
--- a/testdata/src/grouping/errormessage/errormessage-assignment-flow.go
+++ b/testdata/src/grouping/errormessage/errormessage-assignment-flow.go
@@ -115,7 +115,7 @@ func test11(s []*int) {
 	print(*y)
 }
 
-func test12(mp map[int]S, i int) {
+func test12(mp map[int]*S, i int) {
 	x := mp[i] // unrelated assignment, should not be printed in the error message
 	_ = x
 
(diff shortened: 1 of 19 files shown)


testdata/src/grouping/errormessage/errormessage-assignment-flow.go
```diff
@@ -115,15 +115,15 @@ func test11(s []*int) {
 	print(*y)
 }
 
-func test12(mp map[int]S, i int) {
+func test12(mp map[int]*S, i int) {
```
yuxincs, line 118: Can we add test cases where there shouldn't be a diagnostic for missing key for a map with non-nilable element?

util/typeshelper/typeshelper.go
```diff
@@ -108,6 +108,18 @@ func IsDeeplyType[T types.Type](t types.Type) bool {
 	})
 }
 
+// IsDeeplyTypeWithNilableElement is like IsDeeplyType, but additionally requires the element type
+// of T to admit nil.
+func IsDeeplyTypeWithNilableElement[T interface {
```
yuxincs, line 113: Nit: I feel having a HasNilableElement such that you can easily compose IsDeeply && HasNilableElement is better?

But maybe you have already considered that and it's bad for performance?

yuxincs (approved):

testdata/src/grouping/errormessage/errormessage-assignment-flow.go
karthikbhata97: Added test cases

util/typeshelper/typeshelper.go
karthikbhata97: Currently all callers always make both the checks, so keeping it as a single function for now.

PR #516, opened 2026-09-11
chore(coverage): add unit tests for the 'annotation' package

ccuetoh:
## Motivation
Adding unit tests to the 'annotation' package. I'm increasing code coverage for currently untested packages, and this path has no coverage and is easy to unit test.

## Testing
Ran with race detection and 10 iterations to discard flakiness
```
> go test -count=10 -race ./annotation      
ok      go.uber.org/nilaway/annotation  6.936s
```


yuxincs:
I think here since they're not isolated helpers, it would probably be much more valuable if we can add some test code in `testdata` to trigger the logic being tested. Or is it not possible to do so?

PR #521, opened 2026-09-15
Fix false positive for always non-nil error returns across packages

### LINKED ISSUE #365: False positive case
❌ Error case (a.New from another package):

```go
// a/a.go
type A struct {
	V int
}
func New() (*A, error) {
	a := &A{}
	if rand.Int()%2 == 0 {
		return a, errors.New("random error")
	}
	return a, nil
}

// b.go
func test() {
    x, _ := a.New()
    print(x.V) // nilaway reports an error
}
```

✅ No error case (New in the same package):
```go
type A struct {
	V int
}
func New() (*A, error) {
	a := &A{}
	if rand.Int()%2 == 0 {
		return a, errors.New("random error")
	}
	return a, nil
}

func test() {
    x, _ := New()
    print(x.V) // nilaway does NOT report an error
}
```

yamaankhan20:
## Summary

This PR fixes a false positive that occurs when an error-returning function is defined in another package and its non-error result is guaranteed to be non-nil across all return paths.

NilAway can determine that the result is always safe while analyzing the defining package, but that information was not preserved when the result was consumed from another package. As a result, downstream usage could incorrectly report a missing guard.

Fixes #365.

### Removed / Reworked

- Reworked the previous local-only guard-missing filtering logic, which relied on finding both `GuardMissing` and matching return triggers within the same package.
- Replaced the `nonnilCnt`-based check with an explicit `allNonNil` check while analyzing tracked return sites.
- Removed the dependency on having a local `GuardMissing` trigger before an always-safe return site can be identified.

### Added

- Record return sites whose tracked return values are all definitely non-nil.
- Propagate this information across package boundaries through inference facts.
- Add `FalseBecauseAlwaysSafeReturn` to identify non-nil results established specifically by the always-safe return analysis.
- Register the new explanation type for inference fact serialization.
- Use imported always-safe return facts to remove corresponding `GuardMissing` triggers in downstream packages.
- Add a positive cross-package regression case where the result is always non-nil and should not produce a diagnostic.
- Add a negative/control case where the result can be nil and NilAway must continue reporting the unsafe access.

## Testing

Ran the focused error-return tests:

`go test . -run 'TestNilAway/ErrorReturn$' -count=1`

Ran the full test suite:

`go test ./... -count=1`

All tests pass.


yamaankhan20:
@yuxincs And @ccuetoh  kindly review this PR!!!

PR #522, opened 2026-09-18
test: remove repeated fact determinism checks

yuxincs:
Remove the ten-pass fact-encoding determinism check now that `analysistest` checks determinism when inheriting facts ([upstream CL 804360](https://go-review.googlesource.com/c/tools/+/804360)). Retain non-empty encoding, decode/re-encode equality, and fact-size checks for all collected NilAway facts.

Validation: `TestFacts` passes with race detection and coverage on Go 1.26.8.


ccuetoh (approved):

yuxincs:
Well, after some quick local measurement it seems that this doesn't introduce a lot of overhead really (since it only cares about NilAway facts). Having this provides a little extra production. So let's keep this.

PR #525, opened 2026-09-20
chore(ci): pin runners, fix dependency caching, disable fail-fast

ccuetoh:
## Motivation
This PR fixes multiple issues with the GitHub Actions defined for CI.
- ubuntu-latest is migrating to Ubuntu 26 on October 19, 2026. This change pins all workflows to ubuntu-24.04 ahead of that migration so we can upgrade in a controlled way.
- The Test and Integration Test jobs used the cache twice. Once via actions/setup-go and once via a manual actions/cache step, so both matrix legs raced the same cache entry. This surfaced as repeated 'Failed to restore' messages on the cache step. This PR drops the manual steps and uses setup-go's cache instead.
- Disabling fail-fast for the test runs, since it cancelled the sibling runs on a different version.
- Add job-level timeout to stop stuck jobs from running up to 360m.

## Testing
GitHub Actions that ran as part of this PR


.github/workflows/ci.yml
```diff
@@ -8,7 +8,8 @@ on:
 jobs:
   lint:
     name: Lint
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
```
yuxincs, line 12: Do we expect breakages from latest ubuntu? since most of our dependencies are loaded in a controlled way (i.e., using official actions), here we merely need a linux container, such that `ubuntu-latest` would suffice.

yuxincs: Also, any reason we have to give a specific timeout rather than just relying on default?

I'm not opposing having a stricter timeout, just want to better understand here.

.github/workflows/ci.yml
```diff
@@ -40,8 +41,12 @@ jobs:
 
   test:
     name: Test
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
+    timeout-minutes: 30
```
yuxincs, line 45: Same question above

.github/workflows/ci.yml
```diff
@@ -52,12 +57,9 @@ jobs:
...
           cache: true
-
-      - name: Load cached dependencies
-        uses: actions/cache@v6
-        with:
-          path: ~/go/pkg/mod
-          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          cache-dependency-path: |
+            go.sum
+            tools/go.sum
```
yuxincs, line 62: Not sure if it's possible to say `**/go.sum` here 👀 ? this would be future-proof in case we add/drop another module in the future.

.github/workflows/ci.yml
```diff
@@ -68,7 +70,9 @@ jobs:
       - name: Test and generate coverage report
         run: make cover
         env:
-          GOFLAGS: -timeout=20m
+          # Building and running the test binaries is very memory-heavy. Running
+          # them in parallel can overwhelm the runner's memory limit. Serialize them (-p=1).
+          GOFLAGS: -timeout=20m -p=1
```
yuxincs, line 75: Hmmmm, I'm not sure how much `-p=1` helps, because the memory-heaviest test would be the top-level one, which contains many parallel sub tests that may blow up memory. But this helps, of course. We can see if we still hit crashes after this :)

yuxincs:
Nice! Left some questions for me to better understand

PR #526, opened 2026-09-20
feat(ci): shard golden tests to avoid memory exhaustion in ci

ccuetoh:
## Motivation
This PR fixes flakiness in the Golden Test CI runner by introducing a controlled sharding process that prevents memory exhaustion. As of this writing, GitHub Actions runners have 16GB of available memory. These test runs run very close to that limit, and depending on the GC runs, can either lead to an OOO-kill or succeed, leading to indeterministic behaviour and flakiness.

## Testing
PR #524 was used as a test bed for the theory. The added logs there prove the memory exhaustion theory and also that the fix significantly reduces the memory pressure in the test.

The Action run against this PR validates that no results are altered in the report.


yuxincs:
So, I'd propose the alternative:

1. We get this reviewed, tested internally, and merged https://github.com/uber-go/nilaway/pull/503 such that NilAway would behave normally on modular drivers (i.e., transitive facts are propagated normally and won't be dropped).
2. We overhaul the standalone NilAway CLI to wrap around `go vet` instead of `singlechecker` such that the standalone CLI becomes modular (and won't occupy tons of memory).
3. Profit! 

It would offer the following benefits: (1) a lot of complaints from OSS communities around large memory consumption on large projects would be resolved, and we can close several existing issues here, and (2) golden tests on github actions CI should no longer crash, and (3) we won't have to maintain the sharding logic just for a weak CI.

wdyt?

ccuetoh:
I agree in principle, but I have concerns about CI's immediate state. From what I understand, moving to go vet is significant work and is still some time away.

From experience, the CI is failing over half of the runs on OOO-kills, and it's functionally breaking new PRs and stalling contributions. I would suggest we merge this as an interim solution while we work on the overhaul.

yuxincs:
> From what I understand, moving to go vet is significant work and is still some time away.

It's actually pretty straightforward change, the only thing preventing us to do that is because we don't want to break existing behavior for standalone CLI users (because the fact propagation fix is not merged).

With that being said, I'm not opposed to merging this if we remove the sharding logic once the propagation fix is merged and overhaul is done. 

With sharding in place, can we remove `GOMEMLIMIT` hack? Since memory consumption of each shard should be now under github action's memory limit.

